PR #8, opened 2021-07-01
WIP: Packaging and testing

johnomotani:
In looking for how to set up tests in a standard 'Julian' way, I ended up trying to create a 'package', like in the docs here https://pkgdocs.julialang.org/v1/creating-packages/. I'm not super happy with the way I found to do it, because the changes are more drastic than I would like: following the discussion on this thread https://discourse.julialang.org/t/dependencies-of-src-files-inside-a-package/43144/3 (especially the linked comment), I got rid of all the 'internal' modules, and just `include()` the source files in the `moment_kinetics` module in `moment_kinetics.jl`. I should look around more for other documentation/discussion on what 'best practice' for packaging modules is...

This approach does work though, and eliminates some 'boilerplate' code. Setting up as a package also lets `Pkg` handle the dependencies - you add them to the 'project' and then they're tracked in the `Manifest.toml` file. The tests are set up in a 'standard' way, so `Pkg` can run them (so far the only test is the 'fundamental theorem of calculus' test that Michael already set up). I found a Github Action that sets up and tests a Julia package https://github.com/julia-actions/julia-runtest, so the tests now run on Github's CI server, by adding the `.github/workflows/test.yml` file copied from the `README.md` of that action (the test results are shown below).

I should probably add more explanation here, but it's late now - I'll edit later. Comments/questions more than welcome!


johnomotani:
Replaced by #9.

PR #9, opened 2021-07-03
Packaging and testing (new version)

johnomotani:
I learned about submodules from here https://docs.julialang.org/en/v1/manual/modules/#Submodules-and-relative-paths. Structuring the existing modules as submodules of `moment_kinetics` seems much nicer to me than getting rid of them (as in #8), but also removes the need for the `push!(LOAD_PATH, ".")` statements which were (as far as I understand currently) previously recommended, but are now discouraged, in favour of using `include()`.

This PR turns the code into a Julia 'package' and runs the tests automatically through a Github Action on pushes and PRs. Work-in-progress: adding more tests.

The process of making the code into a 'package' involves:
* moving (most of) the source files into `src/`
* creating `Project.toml` and `Manifest.toml` files (the latter auto-generated by `Pkg`, see https://pkgdocs.julialang.org/v1/toml-files/)
* moving tests into `test/`, and setting up so that they run by executing `test/runtests.jl`
* added separate scripts `run_moment_kinetics.jl` and `run_postprocessing.jl` in the top-level which contain/replace the runnable parts of `moment_kinetics.jl` and `postprocessing.jl` (which were moved into the `src/` subdirectory).

Advantages of packaging:
* dependency management is simpler and more maintainable
    * dependencies are now added to the project using `Pkg` and should be automatically installed/updated when the package is activated (see updated `README.md` for slightly more detail).
* standardised, so for example running the testing in Github Actions requires only copy-pasting a standard workflow from https://github.com/julia-actions/julia-runtest
* it should be possible to install `moment_kinetics` into the base Julia environment by using `Pkg.develop()`, but I haven't tried this - for a self contained project like this, the feature is not necessarily useful, it is probably more useful to use the 'local' (not sure the right word for it) project environment, by starting Julia with `julia --project` (activating the environment is done automatically in the `run_moment_kinetics.jl`, `run_post_processing`, `driver.jl` and `postprocessing_driver.jl` scripts, so no extra arguments are needed if using those).

Replaces #8.


mabarnes (approved):

johnomotani:
Forgot to say in the main description - I'd left this as a 'draft' because I'm working on more tests, but it probably makes more sense to merge now (to avoid merge conflicts) and add new tests in a separate PR.

mabarnes:
Agree that it's best to merge now, as I have various changes to source files that I am about to commit.  I think it will be easier for me to make the changes to your merge than for you to make the changes to mine.

PR #14, opened 2021-07-08
Sound wave tests

johnomotani:
Integrated tests using the sound wave decay case.
* Tests calculate mode frequency and growth rate (which is negative as the modes decay for this case), which are compared to:
    * analytical predictions (from F.I. Parra's calculation)
    * saved values (providing a regression test)
* Several different combinations of physical input parameters are tested.
* The set of tests is run for finite-difference and Chebyshev-pseudospectral discretisations, with (for both cases) 0, 1, 2, or 3 moments split out from the kinetic equation.
* As the code is full-f and the test is a linear mode, which is initialised with a small amplitude, the actual precision of the test outputs is not machine precision, but rather roughly phi/delta_phi*(machine precision). This means (I think) that the expected variation due to rounding errors on the outputs is not close to machine precision - the relative tolerance actually used for the regression tests is 2.0e-8.
* A few minor 'infrastructure' changes to support the tests: make more input options settable; make the 'runs' directory settable (so that test output can be saved in a temporary directory); get an absolute rather than relative path to copy `moment_kinetics_input.jl` so that `run_moment_kinetics` can be called from different directories (`Pkg` runs the tests in the `test/` subdirectory).

Even with the resolutions reduced as far as I could get away with, the tests do take a while to run (~50 mins on the Github Actions CI server, or more like 20 mins on our cluster). Probably it would be better to select a small number of these cases to run by default (i.e. for regression testing while working on the code), saving the full set to be run less frequently (e.g. when pull requests are created rather than on every push to every branch). A suggestion for a useful quick subset: one set of physical parameters, with non-zero CX frequency; just no-split and split 3 moments for finite-difference; keep all 4 splitting cases for Chebyshev (these are a little faster than the finite difference cases). That set would cut the run time down to about 70s to 2 min (without/with compilation time on our cluster).

Edit: now updated to skip most of the tests by default, using a macro `@long`. They can be re-enabled by:
* if running tests as a script, by passing `--long` as a command line argument to
    `julia`.
* if running in the REPL, by modifying `ARGS`, e.g.
    ```
    push!(ARGS, "--long")
    ```
* by passing a `test_args` argument to `Pkg.test()`, e.g.
    ```
    Pkg.test(; test_args=["--long"])
    ```
    Note that the semicolon is necessary (not sure why).

Edit 2: now updated CI so `@long` tests should run on merges/commits to `master`. Need to wait for the merge to see if this works...


johnomotani:
I've updated the tests to skip most of the cases by default, using a macro `@long`. They can be re-enabled by:
* if running tests as a script, by passing `--long` as a command line argument to
    `julia`.
* if running in the REPL, by modifying `ARGS`, e.g.
    ```
    push!(ARGS, "--long")
    ```
* by passing a `test_args` argument to `Pkg.test()`, e.g.
    ```
    Pkg.test(; test_args=["--long"])
    ```
    Note that the semicolon is necessary (not sure why).

At the moment, we are just running the shortened set of tests in the Github CI. I think it should be fairly simple to fix this - I am hoping an option will be added to the package we're using for running tests to allow it https://github.com/julia-actions/julia-runtest/issues/47 - if that takes too long I'll try modifying the action myself.

johnomotani:
I made various changes to the routines that fit growth rate and frequency in this PR, but I'm not sure in the end they actually make any difference. I'd forgotten to start with that because the code is full-f and the test case is a small perturbation, the effective precision on e.g. delta_phi is not machine-precision, but something less accurate, so it's reasonable that the best the regression tests can do is agree to ~2e-8 given different rounding errors (e.g. between ubuntu and macOS). I'd actually started out with that value before changing the fitting routines, and ended up back at it in the end.

Just to make a note for when I forget again: part of the reason fitting is tricky is because the sound wave test case has a standing wave, not a travelling wave, so we can't get the mode frequency from the phase velocity. Instead we have to get both frequency and growth rate from time-variation of the amplitude. I changed the post-processing to start by fitting a cosine to delta_phi at each time point, and extracting the amplitude and offset, but in the end (because of the standing wave) I don't think this was an improvement on just using the value of delta_phi at a particular z-index. Tightening some of the fit tolerances, and increasing the maximum allowed iterations, may have made the fitting slightly more reproducible. One thing that I think did actually help was adding a special case to catch zero-frequency modes: if the amplitude of delta_phi does not change sign, we now assume it's a zero-frequency mode and fit just a decaying exponential to it [rather than ~exp(-t)cos(z)], which makes the growth-rate estimation easier and side-steps the difficulty in fitting accurately a very small or zero frequency.

johnomotani:
Just to leave a note somewhere: it's overkill and unnecessary complication for now, but at some point it might be nice to separate out 'unit' and 'integrated' tests - e.g. put them into different subdirectories of `test/` and then have a top-level file that could include all of them automatically, e.g. a `unit_tests.jl` like
```
@testset "unit tests" for test_file in glob("unit/*.jl")
    include(test_file)
end
```

PR #16, opened 2021-07-16
Nonlinear sound wave test

johnomotani:
Modified version of the sound-wave test, using large-amplitude initial perturbations in both density and temperature. Aims to test cases where the vpa grid needs shifting/scaling - cases with upar or upar and ppar moments split out from the distribution function currently fail. Benchmark data is taken from a no-moments-split Chebyshev-discretized run.

Generalises the initial conditions to allow upar and temperature perturbations, and separate phases for each species/variable.

Includes some functions for interpolation that were useful for comparing simulation output to a smaller set of expected data (and allow comparing both finite-difference and Chebyshev grids).

**Edited to add:**
Defines a macro `@testset_skip` which can be used to replace `@testset` it should be followed first by a string giving a reason for skipping the test, then by the usual arguments to `@testset`. The marked testset shows up in the summary as a broken test. If verbose output is enabled, the reason is printed after the testset name.

Adds a new option for tests `--verbose` (or `-v`), which can be passed like `--long` and prints which tests succeeded (failed tests are printed by default).

Tests are now wrapped up in modules. This seems to me at the moment to be the neatest thing to do - it prevents problems due to re-defining functions/variables in `test/setup.jl` when they are imported in multiple tests (they are still imported multiple times, but only within the module contexts, which is OK), and redefinition of functions like `run_test()` which it's nice to be able to use in each test file without worrying about name conflicts.


johnomotani:
I'd say we can almost merge this, and just leave the tests that currently fail being skipped. The one thing I'd like to do is to skip them in a nicer way than just commenting them out. Using `@test_skip @testset ...` works, but it's not as nice as I'd like - I'm hoping someone might have a better solution on Discourse https://discourse.julialang.org/t/skipping-a-whole-testset/65006.

johnomotani:
OK, I think this is ready to merge now. Finding a way to mark tests to be skipped was surprisingly complicated!

Defines a macro `@testset_skip` which can be used to replace `@testset` it should be followed first by a string giving a reason for skipping the test, then by the usual arguments to `@testset`. The marked testset shows up in the summary as a broken test. If verbose output is enabled, the reason is printed after the testset name.

Adds a new option for tests `--verbose` (or `-v`), which can be passed like `--long` and prints which tests succeeded (failed tests are printed by default).

Tests are now wrapped up in modules. This seems to me at the moment to be the neatest thing to do - it prevents problems due to re-defining functions/variables in `test/setup.jl` when they are imported in multiple tests (they are still imported multiple times, but only within the module contexts, which is OK), and redefinition of functions like `run_test()` which it's nice to be able to use in each test file without worrying about name conflicts.

PR #17, opened 2021-07-20
Fix long tests in CI

johnomotani:
Skip a workaround used by the 'julia-runtest' action for the case when the Julia 'General Registry' is not perfectly up to date - requires a file from that action, so rather than copying the file just skip as it doesn't seem to be too important. Also remove `force_latest_compatible_version` argument to `Pkg.test()` which caused an error.


johnomotani:
The long tests passed on a debug version of this branch, so I'm just going to merge. Fingers crossed!

PR #20, opened 2021-07-24
Performance tests and profiling

johnomotani:
Utilities for running performance benchmark tests, and saving and plotting the results.

Uses `BenchmarkTools.jl` to run several repeats of a test and extract some info on run times (min/median/max) and memory usage. Currently one provided test, with similar input to the 'nonlinear sound wave' integrated test (covers 8 cases with different discretization options).

The benchmark test output can be saved (along with some metadata - the current commit hash, machine where the test was run, and date/time) to a git repo (currently at https://github.com/johnomotani/moment_kinetics_performance_results) so that we keep some sort of history of the benchmark numbers to identify performance improvements or regressions. Whether output should be saved to the results repo, and what the machine name is, are set by a `config.toml` file (for the (very simple) format, see `config.toml.example`).

Includes a script to plot the benchmark data (with x-axis being the order of results in the file, which should roughly correspond to date or commit history).

Edit: Added some scripts for profiling (1) memory allocations and (2) run time (sampling profiler). Used these to find and fix a few extra allocations.

This PR is based on the branch in #16 (for no particularly strong reason). The only new changes in this branch are the files under the `performance-tests/` subdirectory.


johnomotani:
Added some scripts for profiling (1) memory allocations and (2) run time (sampling profiler). Used these to find and fix a few extra allocations - for non-moment-split test cases ('test 1' which is finite-difference, and 'test 5' which is Chebyshev), memory allocations are down by an order of magnitude, and run times are down by ~30-50%.
sound_wave

johnomotani:
I've almost finished a merge of #21 into `wall_bc`, so I'm going to merge this and #21 now, then open a new PR for `wall_bc`.

PR #25, opened 2021-08-30
Wall boundary conditions and ionization collisions

johnomotani:
Introduces wall boundary conditions and ionization collisions. Automated tests coming later.


mabarnes (changes requested):
The changes I made to the coefficients were intentional, though probably could benefit from some more checks for periodic BCs, etc.  The reason I made the changes is that these points are at the downwind boundary and third order upwind requires info beyond this boundary.  For the zero or constant BC cases, we only impose the BC at the upwind boundary, so we do not have the information needed beyond the downwind boundary.  As such I took the easy way out and switched to second order upwind here, which does not require info beyond the downwind boundary.  The former implementation was assuming a zero/constant BC at both ends of the domain, giving unphysical spikes at the downwind boundary.

johnomotani:
:+1: makes sense. The 2nd order forms made the tests in a periodic domain fail, so I've put them back with a special case for periodic bcs.

PR #26, opened 2021-08-30
Test wall BCs against analytic solution from Harrison&Thompson

johnomotani:
Test wall boundary conditions by comparing simulation result to analytical steady-state solution from Harrison&Thompson, which uses an artificial particle source proportional to `delta(vpa)` with constant coefficient.

~~Draft because:~~
* ~~At least with the current input parameters, we do not get a smooth solution from the simulation (not sure if the problem is the parameters, or this indicates that I introduced a bug when merging `swap_z-vpa` and `wall_bc`).~~
* ~~My reading of the analytical expression from Michael's August Excalibur report seems not to have a solution for the value of `R_ion` given in the report - maybe I have a normalization wrong, or something more fundamental??~~


johnomotani:
The test now agrees with the analytical solution (for Chebyshev discretization), like in Michael's Excalibur report. I did have to keep `width=0.5` in the source term (instead of `width=0.1` as in the report), or the test would crash.

Running the test with finite-difference does not agree well. It seems like that version develops a large jump at `vpa = 0` at `z = +/- Lz/2`, actually there is a significant level of f all the way up to `vpa = 0` for all `z`. I guess this throws off the phi profile, so that it does not agree well with the analytic solution.

mabarnes:
The `width=0.1` in the report is incorrect: as you observe, it should be `width=0.5`

PR #27, opened 2021-08-30
Regression test with wall boundary conditions

johnomotani:
Regression test with wall boundary conditions and ionization collisions. No analytical solution, so compares output to saved reference values.

Currently draft because, although the simulation reaches a smooth steady state, at intermediate times it looks like there are small-scale wiggles in the z-direction (in phi and density) and the distribution function does not look smooth. Not sure if this is a problem that needs fixing, or just an indication that the initial conditions were a long way from a smooth solution.


johnomotani:
The merging of #26 got a bit confused - it went into `wall_bc` after #25 was merged. I've rebased to include it in this PR now, to merge into `master`.

The Chebyshev case of this test seems OK - there are some wiggles while the simulation settles down to a steady state, but the final state looks reasonable, and agrees fairly well (within rtol=2.e-3) with a higher-resolution version, when interpolated onto a set of sample points (this is tested with some saved output from the higher resolution simulation).

The finite difference case does not seem to work well. The total density keeps growing throughout the simulation, and eventually the simulation develops oscillations in z and then crashes). @mabarnes I think you said the finite difference case was smoother at one point - I guess I might have broken something in the finite difference boundary conditions during the merges (although weirdly the distribution functions at the boundary looked OK - f_ion was pretty smooth, f_neutral has a kink at vpa=0 but still looks reasonable (no oscillations)).

PR #34, opened 2021-11-20
Shared-memory parallelization

### LINKED ISSUE #7: Parallelising and loop performance
I have been looking at documentation for Julia and some packages for thread-based parallelism and other loop optimisations. Opening this issue to keep notes...

* Just sticking the `@threads` macro on a loop seems to work pretty well (test case and results posted in a comment below).
    * variations on putting the work in a function or just writing it out in full seem to make minimal difference
* I didn't find a nice solution for getting threading with array broadcasting. This thread https://github.com/JuliaLang/julia/issues/19777 is a bit old unfortunately. Suggestions from it:
    * `GPUArrays.jl`. Might be interesting in the long run for actual GPU usage. At one point it had a non-GPU implementation that used threads, but that has been removed from the main repo (they still use it for testing).
    * `Strided.jl`. I haven't understood what it's really for yet - some kind of optimisation of array operations. Does support threading of broadcast operations, but for a trivial test case is slower than just using a threaded loop (as this is not its main purpose). Maybe worth looking at again and testing with multi-dimensional array operations.
* This comment https://github.com/Jutho/Strided.jl/issues/9#issuecomment-865898274 suggested looking at `LoopVectorization.jl`. This package is aimed at optimising loops using things like AVX instructions. I ran into a bug trying to use a `tan` function - `ERROR: LoadError: UndefVarError: tan_fast not defined` - and for a trivial test (similar to the one in the comment below but swapping `exp` in place of `tan`) didn't get better performance than just using `@threads`.
* According to the `README.md` of `LoopVectorization.jl` https://github.com/JuliaSIMD/LoopVectorization.jl/blob/master/README.md#broadcasting
    > Note that loops will be faster than broadcasting in general. This is because the behavior of broadcasts is determined by runtime information (i.e., dimensions other than the leading dimension of size 1 will be broadcasted; it is not known which these will be at compile time).

johnomotani:
Adds shared-memory parallelism using MPI. Arrays for distribution function, moments, and a few others are created as MPI shared-memory arrays. Each MPI process then loops over different sets of indices.

Using shared memory means that some places can have an outer loop over z, with inner loop/operation on vpa serial (useful for derivatives), while others have an outer loop over vpa, with the inner loop/operation on z serial. When the loop access pattern changes, need to call `block_synchronize()` (which calls `MPI.Barrier()`) to avoid race conditions.

Also includes a debugging module, so far used mostly to check for race conditions and for redundant calls to `block_synchronize()` (which if they exist should be removed for optimization). Debugging code is only activated if a 'debug level' (set by `-d`/`--debug` command line argument) is set to a high enough value. Macros for this are defined in `debugging.jl` - to define a new macro for adding new debugging features, its name and debug level only need to be added to the `macro_names` list at the beginning of the `debugging` module.

Strong scaling run-times and efficiency shown below (see also #7 for more discussion):
sound_wave_strong_scaling
sound_wave_efficiency

Closes #7.


johnomotani:
I think this is finally ready for review/merge! Spent the rest of last week chasing down a few last bugs and issues with performance test scripts and CI.

johnomotani:
OK, I've been through this PR again today, and fixed a couple of last small bugs. Going to merge now, and fix anything I've missed in a future PR.

PR #41, opened 2022-01-12
Macro expansion examples

johnomotani:
Adds a script that prints all the loop macros created in the `looping` module along with macro-expansions of an example call.

We talked about writing to a file. Could do that easily if people prefer, but having got this far just printing to stdout seemed like a reasonable thing to do: not being saved to a file it's clear that the output is just output and can't be edited to change the code, etc.

Also includes a commit bumping the required Julia version to 1.7.0. Also adds the version requirement to the `Project.toml`, which I'd hoped would result in nicer error messages if someone tries to use a too-old version of Julia, but unfortunately this seems not to be the case: https://discourse.julialang.org/t/strange-error-for-unsupported-julia-version/74489


johnomotani:
@mrhardman does this do what you wanted?

mrhardman:
The script is exactly what I was hoping for. I have tested it on a newly installed version of Julia 1.7.1, and the results are very helpful. Please feel free to merge to the master.

PR #42, opened 2022-01-13
Fix loop macro documentation

johnomotani:
When used to index the example array, the indices printed by `util/print-macros.jl` were in the reverse of the correct order.

Also fix typo in `README.md`.


johnomotani:
Thanks for spotting @mrhardman!

I'll merge this when the tests pass.

PR #51, opened 2022-01-29
Optimize interpolation for mk_wallbc2

johnomotani:
Provides significantly faster interpolation for both Chebyshev and finite difference modes.

Poor performance of the interpolation is the reason the parallel tests were timing out previously.

Chebyshev is numerically identical to previous implementation, just gets rid of allocation of a bunch of temporary arrays that were causing a big slow-down.

Finite difference switches from a cubic B-spline interpolation (which has continuous 2nd derivative, requiring an equation system to be solved), to cubic Hermite spline (calculated using derivatives at the grid points from `derivative!()`, which does not require an equation system to be solved). The new implementation is significantly faster, but possibly a bit less smooth/accurate. Not sure exactly why, but I needed to increase `vpa_L` and `vpa_ngrid` for some of the tests to get them to pass. In particular, the fit to `phi` in the sound-wave tests was getting larger fit errors.


johnomotani:
@mabarnes I've just made this PR to make it easier to see the changes that I'm proposing to merge into `mk_wallbc2-rebased`. I'd suggest merging straight away.

PR #61, opened 2022-03-23
Upar evolution wall bc

mabarnes:
added option to evolve upar as well as density with ionization collisions and wall BC.  Use of the peculiar velocity as a phase space coordinate, as is done for upar evolution, is not recommended for use with the wall BC, given the fact that the wall BC works with dz/dt as a coordinate rather than the peculiar velocity.


johnomotani (approved):
Looks good to me!

PR #62, opened 2022-04-28
Allow `begin_*_region()` calls to be put in more logical places

johnomotani:
Make begin_*_region() return immediately if code already in that region. Intention is to allow begin_*_region() calls to be placed logically (e.g. at the beginning of functions) without having to eliminate redundant calls. This allows the code to be written so that is not necessary to search back through what was called previously to find the most recent begin_*_region() call in order to know what the current region is.

Move `begin_*_region()` calls to more logical places, e.g. `begin_s_r_z_region()` is called at the beginning of
vpa_advection!() to declare that the body of that function is in a `s_r_z` parallel region.

Includes update to the developer docs to describe the new behaviour and recommended style.


mrhardman:
I am happy with these suggested changes, which should clarify the current layout of the data parallelisation in any given low-level subrountine. Has the optional 'no synchronise' feature been tested in place in the code? I (think) am aware that in some languages, optional argument defaults are overwritten by function calls at runtime -- can we make sure that this is not the case in julia?

johnomotani:
Thanks for the review @mrhardman!

The `no_synchronize` argument is used here
https://github.com/mabarnes/moment_kinetics/blob/b591673a4de5a6f536c4a0d166bbf27dff9a8ef0/src/time_advance.jl#L691
and I think it needs to work correctly to make the parallel 'debug checks' pass.

I also checked that julia does not share the awful gotcha/bug that Python has with mutable default arguments. In julia `a` starts as an empty Vector on both calls
```julia
julia> function foo(a=[])
           append!(a, 1)
           println(a)
       end
foo (generic function with 2 methods)

julia> foo()
Any[1]

julia> foo()
Any[1]
```
whereas in Python `a` in the second call retains the thing appended in the first call
```python
>>> def foo(a=[]):
...     a.append(1)
...     print(a)
... 
>>> foo()
[1]
>>> foo()
[1, 1]
```

Plain data types (Booleans, integers, floats, etc.) are fine in either language
```julia
julia> function bar(b=false)
           println(b)
           b = true
           println(b)
       end
bar (generic function with 2 methods)

julia> bar()
false
true

julia> bar()
false
true
```
```python
>>> def bar(b=False):
...     print(b)
...     b = True
...     print(b)
... 
>>> bar()
False
True
>>> bar()
False
True
```

PR #64, opened 2022-06-21
Start writing automated tests for MMS; renormalise manufactured distribution functions

johnomotani:
Start on automated tests using MMS. Time-dependent test `test/manufactured_solutions_tests.jl` is in very rough state. Another test which only tests the 'rhs' evaluation at t=0, `test/manufactured_ddt_tests.jl` is closer to being finished, but still just prints the errors, convergence orders (and a bit of other debugging info) - still needs a decision on what resolution parameters to test automatically, and a pass/fail check and tolerance setting. Note that increasing the resolution too much can easily lead to running out of RAM due to the 5d neutral distribution function.

To get the MMS tests to converge nicely, redefine the manufactured distribution functions so that the ion distribution function when integrated over 0<=vperp<=Lvperp, -Lvpa/2<=vpa<=Lvpa/2 gives exactly the manufactured ion density, and the neutral distribution function integrated on -Lvzeta/2<=vzeta<=Lvzeta/2, -Lvr/2<=vr<=Lvr/2, -Lvz/2<=vz<=Lvz/2 gives exactly the manufactured neutral density. This avoids the error tending to a floor related to the Lv* settings (although this error can be made very small - below machine precision - by increasing Lv* by a factor of 2 or so).

The charge exchange term still only converges at somewhere between 1st and 2nd order (for the resolutions I have tested) - I think this is expected due to the use of linear interpolation in this term. This would probably be improved dramatically by @mrhardman's suggestion of using a vpa,vperp,gyrophase grid for the neutrals.

Fixes a couple of small bugs: copy-paste error in an index in charge_exchange_3V!() function, could have had a significant impact (db2cf67cf66f1589ab0acd3818915daaddbd91eb); wrong (?) region in `ssp_rk!()` may cause errors for some combinations of dimension parallelisation 1803b44f64553f624a96c2a6499ef904d0922770.

### DIFF
--- a/src/manufactured_solns.jl
+++ b/src/manufactured_solns.jl
@@ -92,7 +92,7 @@ using ..type_definitions
     # ion density symbolic function
     function densi_sym(Lr,Lz,r_bc,z_bc)
         if r_bc == "periodic" && z_bc == "periodic"
-            densi = 1.5 +  0.1*(sin(2.0*pi*r/Lr) + sin(2.0*pi*z/Lz))*sin(2.0*pi*t)  
+            densi = 1.5 +  0.1*(sin(2.0*pi*r/Lr) + sin(2.0*pi*z/Lz))*cos(2.0*pi*t)  
         elseif r_bc == "Dirichlet" && z_bc == "periodic"
             #densi = 1.0 +  0.5*sin(2.0*pi*z/Lz)*(r/Lr + 0.5) + 0.2*sin(2.0*pi*r/Lr)*sin(2.0*pi*t)
             #densi = 1.0 +  0.5*sin(2.0*pi*z/Lz)*(r/Lr + 0.5) + sin(2.0*pi*r/Lr)*sin(2.0*pi*t)
--- a/src/manufactured_solns.jl
+++ b/src/manufactured_solns.jl
@@ -325,7 +325,7 @@ using ..type_definitions
                               nr::mk_int, advance::Union{advance_info,Nothing}=nothing)
         rhs_ion_sym, rhs_neutral_sym = manufactured_rhs_sym(Lr,Lz,Lvpa,Lvperp,r_bc,z_bc,composition,geometry,collisions,nr,advance)
         return build_function(rhs_ion_sym, vpa, vperp, z, r, t, expression=Val{false}),
-               build_function(rhs_neutral_sym, vpa, vperp, z, r, t, expression=Val{false})
+               build_function(rhs_neutral_sym, vz, vr, vzeta, z, r, t, expression=Val{false})
     end
 
     function manufactured_sources(Lr::mk_float,Lz::mk_float,Lvpa::mk_float,Lvperp::mk_float,r_bc::String,z_bc::String,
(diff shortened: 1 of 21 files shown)


src/manufactured_solns.jl
```diff
@@ -225,5 +230,38 @@ using Symbolics
...
+            t::mk_float, r::AbstractVector, z::AbstractVector, vperp::AbstractVector,
+            vpa::AbstractVector)
+
+    Create array filled with manufactured solutions.
+
+    Returns
+    -------
+    (densi, phi, dfni)
+    """
+    function manufactured_solutions_as_arrays(
```
mrhardman, line 394: Does a function like this make it possible to avoid passing the MMS functions around the main code?

If I not mistaken, there appears to be a speed cost associated with passing the runtimegenerated function -- it would be interesting to know if you notice a significant speedup by using this approach instead.

johnomotani: I don't think there would be, mostly because this function allocates new arrays for its output, then returns them. Also it's re-building the manufactured-solutions functions each time it's called.

It should be possible to set up a struct in a way that everything in the main time-loop is fully typed. I'll have a look now and check if there is some type instability...

src/manufactured_solns.jl
```diff
@@ -102,12 +102,13 @@ using ..type_definitions
...
-        #if (r_bc == "periodic" && z_bc == "periodic")
-            dfnn = densn * exp( - vpa^2 - vperp^2 )
-        #end
-        return dfnn
+    function gyroaveraged_dfnn_sym(Lr,Lz,Lvzeta,Lvr,Lvz,r_bc,z_bc,geometry,composition)
+        dfnn = dfnn_sym(Lr,Lz,Lvzeta,Lvr,Lvz,r_bc,z_bc,geometry,composition)
+        # Because of particular form of odd component in dfnn_sym, if we set vzeta=-vr
+        # then we eliminate the odd component in the perpendicular velocities,
+        # effectively gyro-averaging
+        gav_dfnn = substitute(dfnn, Dict(vzeta=>vperp/sqrt(two), vr=>-vperp/sqrt(two), vz=>vpa))
```
mrhardman, line 110: I am worried that this operation is only correct when the field line pitch is inifinite (or zero), so that z and b are codirectional. I think that you are wanting vzeta and vr to be the velocities in the plane perpendicular to the magnetic field line, which they are not. In my original 'gyroaveraged' functions code, I was exploiting that vperp^2 + vpa^2 = vtot^2 = vz^2 + vr^2 + vzeta^2, so the gyroaverage was trivial. Am I missing the point here?

mrhardman:
Looking at the manufactured_solns.jl file, I have two immediate comments. You say in the commit message that the need for the boundary terms goes away as soon as the length of the domains are large enough. Since computing the boundary terms is painful and not automated, is it possible to dispense with them by making sure that the tests cover a reasonably large range in velocities? 

I am not sure that the method that you are using for the boundary terms will carry over to the wall boundary test currently in the vanilla branch. For the neutrals, the distribution function must be proportional to the Knudsen distribution in order to match the boundary condition.

Finally, a matter of taste. I rather liked having the kinetic equation and the drift-kinetic equation written on one line, as we would write it for a slide. I thought that this would have value by the way of self documentation. I can see why it could be convenient to add the terms one by one to a "rhs". If we want to go that way I think we need to make sure that there are adequate comments documenting which equation is being solved in the source code near this function. Since Julia permits expressions extended over multiple lines, I would have thought that we could avoid spreading the equation over too much of the source code.

src/manufactured_solns.jl
johnomotani: I made a mistake here - was thinking z and b are in the same direction. Was trying to be clever and give non-zero flow, but this doesn't work :disappointed:

Issue #65, opened 2022-08-18
All MPI controlled processes do not exit with error on a single process


mrhardman:
When running with MPI, I have noticed that if there is an error that causes Julia to abort the calculation on a single process, then the program does not exit all processes. Instead, the MPI processes which have not met an error remain stalled. This may be dependent on the architectural which we run the code.

johnomotani:
Yes, I've noticed the same problem. I've attempted a solution on my 1D1V branch in 7cac085cbe12bc7f3951588fad32622de89f7f43 and 7efef078c5ff86d567ea328cf10aa79209703650. Not sure it works all the time though...

mrhardman:
Potentially related to https://github.com/mabarnes/moment_kinetics/issues/86. Closing as stale, can re-open if similar issues are reported.

PR #67, opened 2022-08-22
Better debug checks for parallelisation and some bug-fixes

johnomotani:
Previously, the debug checks for correctness of parallelisation did not work in the 2D2V branch as they did not use enough processors on the CI server. I've also just realised that it was possible for bugs (race conditions) to slip through the checks, if the processor-splitting for different region types happened to be the same (e.g. an s_r_z region and an s_r_z_vpa region, where actually vpa was not parallelised and so the reads/writes would happen at the same array locations on each process).

This PR introduces a much more robust (but slower) set of correctness checks. A special debug initialisation function for the loop ranges is used. Now the checks are run repeatedly. On each repeat just one type of region is parallelised, so that any conflicts are guaranteed to show up as all the other regions are run in serial-style (only accessing on process-0). In addition, the number of processes is divided up as much as possible (e.g. into 8=2*2*2) to split as many dimensions as possible, but if there are not enough processes, then extra repeats are used so that every dimension in the given region type is actually split in one run.

Adds a debug check with MMS functions being used.

The debug check CI run is split into several jobs so that the various tests can be run in parallel, as they all now take quite a long time.

Includes various bug fixes for the shared-memory parallelism that were found while improving the checks.


johnomotani:
The tests seem to be timing out. Not sure if that's Github's servers running more slowly than they used to, or whether something I did has slowed the tests down. I should do some performance benchmarking at some point to figure it out (at least if re-running the tests tomorrow doesn't help).

johnomotani:
By the way, using `--compiled-modules=no` seems to break `Symbolics.jl` at the moment (https://github.com/JuliaSymbolics/Symbolics.jl/issues/706). This means that the 'normal' way of activating debug features doesn't work. The best/only workaround I have so far is to just edit `debugging.jl` and put the (integer) value you want in for `_debug_level`
https://github.com/mabarnes/moment_kinetics/blob/ec93f59ffd64846cb8ef1ff6928c7b720f67f3f7/src/debugging.jl#L46

mrhardman:
Do these changes include the fix to

https://github.com/mabarnes/moment_kinetics/issues/65

I have not currently understood your solution to issue #65 well enough to merge!

johnomotani:
> Do these changes include the fix to
#65

No they don't. I can make another PR for that!

johnomotani:
@mrhardman this PR is now based on `radial-vperp-standard-DKE-with-neutrals-Julia-1.7.2`, and I think is ready to merge.

PR #68, opened 2022-08-23
Stop all MPI processes when an error occurs

### LINKED ISSUE #65: All MPI controlled processes do not exit with error on a single process

johnomotani:
If just one process throws an exception, we need to catch it and call `MPI.Abort()` to stop all processes.

Also uses a `try...finally...end` construct to try to call `cleanup_moment_kinetics!()` when an error occurs, to close output files properly, etc. This can probably fail if `MPI.Abort()` is called as I guess `MPI.Abort()` kills the other processes without waiting for exception handling. Unfortunately I don't know of a decent solution for error/exception handling with MPI...

Fixes #65.

### DIFF
--- a/src/moment_kinetics.jl
+++ b/src/moment_kinetics.jl
@@ -75,24 +75,33 @@ using .time_advance: setup_time_advance!, time_advance!
 main function that contains all of the content of the program
 """
 function run_moment_kinetics(to::TimerOutput, input_dict=Dict())
-    # set up all the structs, etc. needed for a run
-    mk_state = setup_moment_kinetics(input_dict)
+    try
+        # set up all the structs, etc. needed for a run
+        mk_state = setup_moment_kinetics(input_dict)
 
-    # solve the 1+1D kinetic equation to advance f in time by nstep time steps
-    if run_type == performance_test
-        @timeit to "time_advance" time_advance!(mk_state...)
-    else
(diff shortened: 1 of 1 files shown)


mrhardman:
I have checked the behaviour of this code in a case with a known error, and indeed the cores do exit. I am happy for the merge to go ahead. 

It would be useful to improve the error messages, but otherwise, this is a great improvement.

src/moment_kinetics.jl
```diff
@@ -75,24 +75,38 @@ using .time_advance: setup_time_advance!, time_advance!
...
+        if block_rank[] == 0 && run_type == performance_test
+            # Print the timing information if this is a performance test
+            display(to)
+            println()
+        end
+    catch e
+        # Stop code from hanging when running on multiple processes if only one of them
+        # throws an error
+        if global_size[] > 1
+            println(e)
```
johnomotani, line 105: > It would be useful to improve the error messages

Something like this?
```suggestion
            println("Abort called on rank $(block_rank[]) due to error. Error message "
                    * "was:\n ", e)
```

PR #71, opened 2022-09-04
Enable restarting of runs

johnomotani:
Add a function `restart_moment_kinetics()` that allows restarting a run from any time point in an existing output file.

Save the normalized pdf rather than an unnormalized one to output files - makes restarting simpler as the normalized pdf is the evolved variable.

Call MPI.Abort() if an error occurs on any process, printing the error message first. This helps to avoid a run hanging if an error occurs on only one (or a few) processes.


johnomotani:
Should really add a test for the restart functionality, but I think we can leave that for later...

PR #74, opened 2022-09-10
Initial conditions for wall bc cases

johnomotani:
New setup for initial conditions in wall boundary condition cases:
* Make distribution functions go to zero smoothly where (lab-frame) v_parallel goes to 0.
* For moment-kinetic versions first initialise `f` as full-f, then calculate moments from it, and use them to normalise `f` and shift to the w_parallel grid.
* Construct initial `f` starting at the wall boundary grid points, so that it matches the boundary conditions, then build `f` in the rest of the grid as a superposition of f(vpa) from the boundaries, adding a scale factor so that the density is not constant.

Tweak settings to fix the run used for precompilation.

Increase the timeout for parallel tests so that the job completes.


mrhardman:
I approve of this method of initialising the PDF where one starts from the unnormalised PDF (perhaps just because it is the method that I took for the MMS tests in the standard DKE implementation for the 2D case. I note that the approach here seems to be at odds with the restarting feature in https://github.com/mabarnes/moment_kinetics/pull/71 where the normalised PDF is stored. Are these changes consistent?

johnomotani:
> I note that the approach here seems to be at odds with the restarting feature in #71 where the normalised PDF is stored. Are these changes consistent?

They are consistent. The implementation is not the most efficient, but when restarting, the code just overwrites all the arrays (after initialisation is completed) with the values read in from the old output file. So when starting at t=0 we create an unnormalized PDF, then normalize and interpolate, but we save the normalized PDF and when restarting we reload it directly (as if the simulation had kept running). It would be more efficient to skip the 'initialisation' entirely when restarting, but I thought it kept the code simpler to avoid the extra `if`-statement.

johnomotani:
On second thoughts, probably the best thing to do is to separate allocating the arrays from initialising them. Then we'd just do something roughly like
```
... = allocate_arrays()
if restarting
    reload!(...)
else
    initialize!(...)
end
```
I might add that on here now... I don't think it should clash with the following PRs.

PR #75, opened 2022-09-10
Second derivative and numerical dissipation

johnomotani:
Introduces a collection of functions to add numerical dissipation, all disabled by default. Currently, these seem to be useful to avoid a numerical (?) instability in the initial stages of wall-bc runs.

The dissipation is switched on and controlled by input parameters. These parameters are passed in a section `[numerical_dissipation]` in the TOML files.

Adds a second derivative function to be used for numerical dissipation. Finite difference version is standard second order centred difference. The Chebyshev pseudospectral version is a tentative attempt at an implementation; the current implementation does the following:
* Take the first derivative, storing the values in each separate element
* Get the continuous first derivative at element boundaries by averaging the values from both elements
* Take the derivative of the continuous first derivative to get a second derivative
* At each element boundary, the first derivative will generally be discontinuous, and the second derivative calculated so far will not necessarily tend to smooth out the discontinuity (or at least not that strongly). To try and help this, add an extra contribution at the element boundaries proportional to the discontinuity in the first derivative between elements.

### DIFF
--- a/docs/src/numerical_dissipation.md
+++ b/docs/src/numerical_dissipation.md
@@ -1,0 +1,6 @@
+`numerical_dissipation`
+=================
+
+```@autodocs
+Modules = [moment_kinetics.numerical_dissipation]
+```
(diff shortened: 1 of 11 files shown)


mrhardman:
Do you think that the changes here may be useful for the standard 2D DK case?

johnomotani:
> Do you think that the changes here may be useful for the standard 2D DK case?

I don't know - maybe?

PR #79, opened 2022-09-27
Ionization fixes

johnomotani:
* `ionization_collisions!()` uses `fvec.dens` and `fvec.upar` instead of `moments.dens` and `moments.upar` as the `moments` versions are only updated at the end of the full timestep, not at each RK stage.
* fix index error in interpolation for evolving_upar case in `ionization_collisions!()`
* add missing contributions like (u[s]-u[s'])^2 to ionization terms in energy equation, and in places where energy equation contributes to `vpa_advection!()` and `source_terms!()`.

With these fixes, I have a 'full moment-kinetic' case (evolving density, upar and ppar) with wall boundary conditions that runs and agrees pretty well with the 'full-f' and 'evolving upar' cases!

### DIFF
--- a/examples/wall-bc/wall-bc_cheb_split1.toml
+++ b/examples/wall-bc/wall-bc_cheb_split1.toml
@@ -1,11 +1,12 @@
 n_ion_species = 1
 n_neutral_species = 1
 boltzmann_electron_response = true
-run_name = "wall-bc_cheb"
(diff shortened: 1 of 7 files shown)


johnomotani:
This PR now includes example input files for the wall-bc case where I got agreement between full-f, evolving-upar and evolving-ppar cases.

PR #81, opened 2022-09-28
Update documentation

johnomotani:
Update documentation with:
* [x] notes on enforcing moment constraints on normalized particle distribution function
* [x] notes on applying boundary conditions
* [x] derivation of equations with all ionization terms included

Also includes adds the `moment_constraints` module to the docs, which was missed in #72.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -26,4 +26,5 @@ jobs:
       - name: Test examples
         run: |
           pip3 install --user matplotlib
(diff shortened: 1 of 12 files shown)


johnomotani:
Unfortunately it turns out that Documenter.jl doesn't seem to support references to equations at the moment. It seems to have switched from MathJax to KaTeX since the comment in this issue https://github.com/JuliaDocs/Documenter.jl/issues/975 which suggested references should work. KaTeX does not currently support references https://github.com/KaTeX/KaTeX/issues/2003.

johnomotani:
Equation numbering looks a bit dodgy at the moment. Don't think it matters too much right now though, hopefully Documenter.jl (or its dependencies) will fix the issue at some point.

Issue #85, opened 2022-10-20
Diagnosis of 2D wall BC MMS test issue


mrhardman:
The MMS test suite is designed to allow for rapid development of new features using a symbolic algebra package to compute the sources in the DKE given a 'target' manufactured solution. For the 2D-3V version of the code with E_r =/= 0 and wall boundary conditions, the test has identified a possible issue of non-convergence with increasing resolution

Here I start to try to track down the error. I start with commit ecfa0f87bc391675b33133eb4a4c68acf20e1e4c. Running the input files
 `runs/2D-wall_cheb-with-neutrals_nel_r_1_z_2_vpa_2_vperp_2`
`runs/2D-wall_cheb-with-neutrals_nel_r_1_z_4_vpa_4_vperp_4`
`runs/2D-wall_cheb-with-neutrals_nel_r_1_z_8_vpa_8_vperp_8`
`runs/2D-wall_cheb-with-neutrals_nel_r_1_z_16_vpa_16_vperp_16`
where I increase the number of elements of all the z, v||, vperp, vz, vr, vzeta grids (for a single r point with E_r = 0) and reduce the time step in tandem, I generate the following figures (with `run_MMS_test.jl`).
[MMS_test_velocity_nelement_1D-3V-wall_cheb-updated.pdf](https://github.com/mabarnes/moment_kinetics/files/9831423/MMS_test_velocity_nelement_1D-3V-wall_cheb-updated.pdf)
[MMS_test_velocity_nelement_1D-3V-wall_cheb-updated_fields_no_Er.pdf](https://github.com/mabarnes/moment_kinetics/files/9831424/MMS_test_velocity_nelement_1D-3V-wall_cheb-updated_fields_no_Er.pdf)

These figures demonstrate declining errors in all fields, approaching machine precision. The error in the neutral density remains surprisingly high.

mrhardman:
I have now generated a series of 2D-3V runs with a spatially 2D grid, and E_r = 0 on the wall boundaries (using commit c6867a051a4a4aa82604ca22b4b4701f561b3166 and the input files from commit bd7fd4bfc49557999f7ead27c845a9d3fb6186bf). Using a similar procedure to the 1D case, I scan in n_element of each coordinate, plus I reduce the time step at fixed (physical time) simulation length. The error fields between the numerical and symbolic moments, dfns and fields are shown below.

[...]

PR #87, opened 2022-10-26
Fix some MPI errors

### LINKED ISSUE #86: 2D code appears to hang without error when running on HPC
Using the branch radial-vperp-standard-DKE-with-neutrals-pdf-animations (commit da276a4df051bbbee3913c47a7a58f17fde1529c) for simulations on a HPC cluster, I noted that the simulation appeared to hang after a relatively small number of timesteps, with no error messages. Using the new screen printout that shows the time when the last write to file was completed, the simulation failed to advance to the next nwrite steps after many multiples of the time that the simulation should have taken. The simulation could only be cancelled. 

e.g.

     beginning time advance...14:27:41
     finished time step 100  14:32:35
     finished time step 200  14:37:25
 
But the simulation ran for another hour without further output of any kind.

johnomotani:
Fix #86:
* The export of `global_size` from the `communication` module was missing, which caused an undefined-variable error in the `catch`-block here https://github.com/mabarnes/moment_kinetics/blob/7975a26ce44f790f2c53e44f34a5ba079c347e74/src/moment_kinetics.jl#L101-L110 that prevented `MPI.Abort()` from being called.
* `MPI.free()` is a collective call, and previously there was a `try...finally...` that tried to call `cleanup_moment_kinetics!()` (which calls `MPI.free()`) on any error. When running on multiple processes, this meant the code would hang if an error occured on some processes and not on others. Removing the `try...finally...` allows errors to trigger `MPI.Abort()` as intended.

Fix inconsistency of some errors reported in #85
* Shared-memory bug, need to use `@serial_region` in `init_knudsen_cosine()`.

Copy fix from 1D1V branch for error in file I/O when using DebugMPISharedArray for debugging.

Nicer printing of caught errors before calling `MPI.Abort()` - may help to diagnose the origin of the errors.

### DIFF
--- a/src/communication.jl
+++ b/src/communication.jl
@@ -12,8 +12,8 @@ loop ranges), as at the moment we only run with 1 ion species and 1 neutral spec
 """
 module communication
 
-export allocate_shared, block_rank, block_size, comm_block,
-       comm_world, finalize_comms!, initialize_comms!, global_rank, MPISharedArray
+export allocate_shared, block_rank, block_size, comm_block, comm_world, finalize_comms!,
+       initialize_comms!, global_rank, global_size, MPISharedArray
 
 using MPI
 using SHA
--- a/src/file_io.jl
+++ b/src/file_io.jl
@@ -507,27 +507,27 @@ function write_data_to_binary(ff, ff_neutral, moments, fields, t, n_ion_species,
         # add the time for this time slice to the netcdf file
         cdf.time[t_idx] = t
         # add the distribution function data at this time slice to the netcdf file
-        cdf.f[:,:,:,:,:,t_idx] = ff
+        cdf.f[:,:,:,:,:,t_idx] .= ff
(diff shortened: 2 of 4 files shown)


mrhardman:
I have tried these changes in https://github.com/mabarnes/moment_kinetics/commit/a04530e622cae4f78ca209dfe81aa95553683eac but as you noted https://github.com/mabarnes/moment_kinetics/issues/86#issuecomment-1292644829 this fix by itself does not fix #86.

johnomotani:
I think I found and fixed the actual bug now! `MPI.free()` is a collective call, and previously there was a `try...finally...` that tried to call `cleanup_moment_kinetics!()` (which calls `MPI.free()`) on any error. When running on multiple processes, this meant the code would hang if an error occured on some processes and not on others. Removing the `try...finally...` allows errors to trigger `MPI.Abort()` as intended.

PR #89, opened 2022-11-03
Fix slowdown of NetCDF file writing

johnomotani:
Pass array references when writing to NetCDF file. Apparently broadcasting the write operation is significantly slow. Instead pass the Array directly (as used to be done), but with a special work-around for DebugMPISharedArray which avoids (forbidden) implicit conversion to Array.

### DIFF
--- a/src/file_io.jl
+++ b/src/file_io.jl
@@ -10,6 +10,7 @@ export write_data_to_binary
 
 using NCDatasets
 using ..communication: _block_synchronize
+using ..debugging
 using ..looping
 using ..moment_kinetics_structs: scratch_pdf, em_fields_struct
 using ..type_definitions: mk_float, mk_int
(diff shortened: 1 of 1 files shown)


johnomotani:
I'd changed, e.g. `cdf.f[:,:,:,:,:,t_idx] = ff` to `cdf.f[:,:,:,:,:,t_idx] .= ff` to avoid problems with `DebugMPISharedArray` not being `convert`-able to `Array`. Unfortunately, it seems like the `.=` form slows things down a lot. I suspect that by broadcasting the assignment operation, it makes a separate 'write to NetCDF' call for each element in the array and this is slow compared to a 'write this whole array at once to NetCDF' call. For the standard, non-debug case, this PR just reverts the change. For the debug case, there is a special variant of the `write_data_to_binary()` function that grabs the `Array` member of the `DebugMPISharedArray` struct directly.

PR #90, opened 2022-12-09
HDF5 I/O for 2D branch

johnomotani:
Refactor `file_io.jl` and `load_data.jl` to use a generic interface to an I/O backend. I/O backends now supported are NCDatasets.jl and HDF5.jl. HDF5 is used by default, can be switched by setting
```
[output]
binary_format = netcdf
```


johnomotani:
@mrhardman this seems to work now - I was able to run and post-process a distributed+shared memory run without running into any exceptions anyway. Can we merge now?

mrhardman:
I am happy for the merge to go ahead. There may now be some extra changes required to keep up with some minor changes in the -mpi branch. I have re-enabled the MMS test plotting script, so there may also have to be upgrades locally there. Happy to take a look at the merge in the new year.

johnomotani:
I've resolved the merge conflicts. The 1d tests pass, and the MMS test case I was checking (that includes wall bc) gives the same results after the merge as before, so hopefully I haven't broken anything. Going to merge now.

PR #92, opened 2023-03-27
Remove Manifest.toml files from git repo

johnomotani:
I suggest we remove the `Manifest.toml` files from the repo. It does not seem to be helpful to track them. They can be generated easily by running `pkg> instantiate`. Often recently I've found myself 'solving' package conflicts or library errors by deleting `Manifest.toml` and re-generating it, so I think this way is probably more robust than having a shared `Manifest.toml` which may break things if, for example, you use a different Julia version than the one that generated `Manifest.toml`.


mabarnes:
agree.  i've had the same issue with having to delete and regenerate, so having the generation of the file as part of the installation instructions seems sensible.

johnomotani:
This PR has been suffering from test 'failures' in the tests that use MPI on macOS. I don't understand the cause, but keeping Julia at v1.7 and preventing Plots.jl from updating from v1.24.3 (which works) to v1.38.8 seems to avoid the problem.
I'm not too worried about this as the tests don't actually fail - they all pass, but then (the ones using MPI on macOS) hang at the point where they should return from the `@testset` block. If you were running interactively you could just Ctrl-C to stop the program and everything would be fine.

PR #94, opened 2023-04-04
Parallel I/O

johnomotani:
Implementation of parallel I/O for the HDF5 output interface.

* Parallel output is now the default if it is supported by the back-end.
* Parallel I/O requires the linked HDF5 library to have been compiled with MPI - the same MPI as is used to run `moment_kinetics`. See notes in `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,23 @@ this can be suppressed by setting
     export QT_QPA_PLATFORM=offscreen
     ```
 in your `.bashrc` or `.bash_profile` files. 
-    
+
+## Parallel I/O
+
+Note that to enable parallel I/O, you need to get HDF5.jl to use the system
+HDF5 library (which must be MPI-enabled and compiled using the same MPI as you
+run Julia with). To do this (see [the HDF5.jl
+docs](https://juliaio.github.io/HDF5.jl/stable/#Using-custom-or-system-provided-HDF5-binaries))
+run (with the `moment_kinetics` project activated in Julia)
+```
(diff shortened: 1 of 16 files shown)


johnomotani:
@mrhardman I've gone back to the parallel I/O implementation. I think it works now! But I haven't tested extensively. If you could test a few of your cases that would be great. The intention is that both running simulations and post-processing should 'just work'.

A couple of things to be aware of:
* When using parallel I/O the output file names are `<run_name>.moments.h5` and `<run_name>.dfns.h5`, without the `irank` label.
* If multiple output files are present in a directory (e.g. `foo.moments.h5` and `foo.0.moments.h5`) the `open_readonly_output_file()` function will throw an error because it doesn't know which to read. This will happen if you re-run, with parallel I/O, a simulation that already had output files. To get around this move or delete the old output files.

I'm proposing to merge the 2D branch with this parallel I/O into the 1D branch, because I think it might make merging the post-processing and restart functionality a little bit simpler.

mrhardman:
I have tested a quick manufactured solutions input file on a Linux server and everything appears to work out of the box, apart from having to set the MPI preferences. Is it possible to revert to the distributed i/o for h5 output if HDF5 cannot be recompiled on the local machine? I tested with netcdf binary output, and all default manufactured solutions test diagnostics also worked there.

I am happy for the merge to go ahead.

johnomotani:
Yes, there's an `parallel_io` option in the `[output]` section that can be used to switch between serial and parallel I/O. Also, the default will switch back to serial if the linked HDF5 doesn't have MPI.

Not sure what we can do about the MPIPreferences thing. Can have a think about that at some point...

PR #96, opened 2023-06-07
2D and MMS

johnomotani:
Add 2D, 'full-f' model (no moment-kinetic version yet), by @mrhardman.

Introduce radial dimension, and extend velocity space to v-parallel and v-perp for ions, and vx, vy, vz for neutrals.

Implements functionality for 'method of manufactured solutions' (MMS) tests. Allows symbolically specifying an 'manufactured solution' and, by evaluating the kinetic equation symbolically, the extra source terms that force that to be a solution to the kinetic equation.

Add distributed-memory MPI parallelism, so the code can now run on multiple nodes of an HPC cluster.

Some extra small features are added after merging in the 2D branch:
* 'Debug checks' which look for shared-memory errors are now much faster, and run properly on the CI tests. It turned out that a lot of the run time was spent in creating stacktraces, which are needed to debug an error, but not to for just testing if there is one in a certain case.
* In the `pdf` sub-structs, the `unnorm` array has been removed as it ended up never being used or needed. The naming is now slightly confusing as the pdf is now stored in a field called `norm`, but this can be tidied up later.
* Moments are written to the `.dfns` output files for distribution functions (as well as the `.moments` output files for moments). This allows the `.dfns` files to be used for restarting simulations, without having to worry about whether the moments were saved at the same time point in `.moments`.
* File I/O now supports `Bool` and `String` types (for single, non-time-dependent variables).
* File names for 'distributed' runs that don't use parallel I/O were changed. Using the naming convention `<run_name>.dfns.0.h5`, etc., makes the file names easier to parse as we can detect whether parallel I/O was used by checking if the first thing before the `.h5` is an integer or `.dfns`/`.moments` (avoiding the issue that a user-specified `<run_name>` might end with `...42.0` or something).
* The input settings (as stored in the `scan_input` Dict) are written into output files, so that they can be read back in in post-processing without needing the input `.toml` file in addition to the output file.


mrhardman:
> I realised the 2D cases I checked didn't include charge exchange or ionization, so also tested `runs/2D-sound-wave_cheb_cxiz_nel_r_2_z_2_vpa_4_vperp_4/`, which again looks OK. Most errors are ~1e-12, the worst in the neutral distribution function which has a maximum absolute error of 1.7e-10.

When you say errors, do you mean differences between the 2D branch and the 1D1V-merged-2D branch? Those errors are much smaller than what I saw in my MMS tests, for example.

In my most recent 2D2V commits I have computed correctly the pressure using easily testable integration routines that I would like to see supported going forward. These routines make it easy to test collision operators in a sandbox. I also included ppar, pperp, upar in the MMS tests for the Wall BC case, proving that my new pressure calculation is correct.

johnomotani:
> When you say errors, do you mean differences between the 2D branch and the 1D1V-merged-2D branch? Those errors are much smaller than what I saw in my MMS tests, for example.

Yes, I just mean the difference between the pre-merge and post-merge branches.

> In my most recent 2D2V commits I have computed correctly the pressure using easily testable integration routines that I would like to see supported going forward. These routines make it easy to test collision operators in a sandbox. I also included ppar, pperp, upar in the MMS tests for the Wall BC case, proving that my new pressure calculation is correct.

Suggest we merge this, then add more new features as self-contained PRs. Your new versions do sound like a good plan!

mrhardman:
> Suggest we merge this, then add more new features as self-contained PRs. Your new versions do sound like a good plan!

Yes, once you have got your current merge underway, then we can talk about which features from my current development should be ported over in separate PRs.

johnomotani:
I think this PR is ready to go. Is there anything else anyone thinks we should check, etc.?

mrhardman:
Have you checked the MMS tests for the Wall BC & Periodic BC cases? Have you checked that the MPI works in multinode cases?

johnomotani:
> Have you checked the MMS tests for the Wall BC & Periodic BC cases?

Sorry, I haven't kept up with all your reports - what cases need running for this, and what do you want to check to call the test 'successful'?

> Have you checked that the MPI works in multinode cases?

I haven't actually checked multiple nodes, but I've set `z_nelement_local` and `r_nelement_local` so that z_nelement_global/z_nelement_local=2, and r_nelement_global/r_nelement_local=2, and the pre- and post-merge agree in that case.

mrhardman:
[MMS_test_vpaz_nelement_1D-1V-wall_cheb_fields_and_ion_pdf_no_Er.pdf](https://github.com/mabarnes/moment_kinetics/files/11679779/MMS_test_vpaz_nelement_1D-1V-wall_cheb_fields_and_ion_pdf_no_Er.pdf)
[figure2.zip](https://github.com/mabarnes/moment_kinetics/files/11679785/figure2.zip)

Figure 2 from report 2070839-TN-04

mrhardman:
[MMS_test_vpazr_nelement0.25_2D-1V-wall_cheb_fields_and_ion_pdf.pdf](https://github.com/mabarnes/moment_kinetics/files/11679799/MMS_test_vpazr_nelement0.25_2D-1V-wall_cheb_fields_and_ion_pdf.pdf)
[figure10.zip](https://github.com/mabarnes/moment_kinetics/files/11679803/figure10.zip)
Figure 10 from report 2070839-TN-04

mrhardman:
[MMS_test_nelement_2D-sound-wave_cheb_fields.pdf](https://github.com/mabarnes/moment_kinetics/files/11679812/MMS_test_nelement_2D-sound-wave_cheb_fields.pdf)
[2D-sound-wave cheb nel r 16 z 16 vpa 16 vperp 16.zip](https://github.com/mabarnes/moment_kinetics/files/11679828/2D-sound-wave.cheb.nel.r.16.z.16.vpa.16.vperp.16.zip)
(you'll have to generate the other points by varying nelement
Figure 5 from 2070839-TN-02

mrhardman:
Are these tests clear? They could be significantly improved by plotting the expected scaling on the resulting figures. I am happy to get on a call to clarify any details regarding running the simulations.

johnomotani:
I think @mrhardman might be on holiday now, so I'm going to declare the tests passed and merge this so I can open some small PRs porting in a useful feature or two from the branch I was using for instability-in-double-periodic-box simulations.

PR #99, opened 2023-06-21
2D instability setup

johnomotani:
Some small changes for running simulations looking for linear instability in a 2D periodic box. Maybe useful to merge now because the updates include some minor fixes and improvements:
* bugfix for `analyse_and_plot()` when there are no neutral species
* make plots/animations of parallel temperature
* option to load only every n'th time point for post processing - can help post-processing large output files if memory is a limitation
* exception handling in post-processing so that (most of the time...) if there is an error making one plot, the other plots are still generated
* animations of the distribution function are made when reading from 'distributed' output files (i.e. distributed memory, without parallel I/O). To implement this, added a function to load some slice of the distribution function from the output files (useful when the full distribution function would be too big to fit in memory).
* when running post-processing, auto-detects whether a run is 1D1V or higher-dimensional, removing the need for hard-coded switches in `post_processing.jl`
* fixes a factor of sqrt(2) in `init_vth!()`. Before this fix, the actual temperature initialised was half the value given in the input file.
* nicer way to get settings for the manufactured solutions code. There is now a `manufactured_solns` section in the input file, which includes a `type` option that can be used to switch between different manufactured solutions.
    * This update is also a prototype for a new way of specifying options and default values, where they only need to be defined in a single place, see https://github.com/mabarnes/moment_kinetics/blob/88484fbde4fa69e3554dd5155300fbc22ea448d6/src/moment_kinetics_input.jl#L257-L278


mabarnes:
I have a question about the change to init_vth: to what extent does the temperature itself actually enter the calculation?  For electrons, it enters explicitly via T_e in the Boltzmann response but for the ions it's not obvious to me that it is used at all in the solution of the equations (since we have formulated everything in terms of ppar and vth).  I think the normalisation of T_e appearing in the Boltzmann response is implicitly T_ref = (1/2) * m_ref * c_ref^2 rather than m_ref * c_ref^2, as otherwise there would be a factor of 2 appearing explicitly in the expression for phi.  vth is normalised by c_s (= c_ref in the absence of electrons), so by my calculation the normalised T should be the square of the normalised vth, as was originally present in the code.  What may be wrong is diagnostics if they exist (but hopefully not actually in the code?) because the normalised ppar should then be ppar = dens * temp * 0.5.  

Do others agree?

The factors of 0.5 flying around are irritating, and are due to my initial decision to normalise pppar by m_r * c_r^2 * n_r, but for T_r = (1/2) m_r * c_r^2.  I'm open to changing one of these norm choices if you guys want...

johnomotani:
It's true temperature is never used, only vth and ppar. We always (I think) have ppar=(1/2)\*n\*vth^2, see
https://github.com/mabarnes/moment_kinetics/blob/72ee32ed6691e14f553938e4a66c626955aa9218/src/time_advance.jl#L1144
https://github.com/mabarnes/moment_kinetics/blob/72ee32ed6691e14f553938e4a66c626955aa9218/src/initial_conditions.jl#L159

I was assuming that ppar=n\*Tpar (rather than as @mabarnes said ppar=1/2\*n\*Tpar, which I agree would be consistent with the code as it was), and then to be consistent with the ppar formula above Tpar=(1/2)\*vth^2, when saying that `init_vth!()` 'should have the factor of 2', and when adding plots of 'temperature' here
https://github.com/mabarnes/moment_kinetics/blob/88484fbde4fa69e3554dd5155300fbc22ea448d6/src/post_processing.jl#L2093

In the report "Numerical study of 1+1D drift kinetic models for parallel dynamics" (I've got it labelled as TN-10), Table 1 on p. 15 @mabarnes wrote explicitly that ppar is normalised by m\_s\*N\_e\*c\_s^2, and Tpar is normalised (in the way that agrees with what I was saying in the previous paragraph) by m\_s\*c\_s^2.

I agree phi being normalised by T_e/e is consistent with ppar=1/2\*n\*Tpar and not with my change to `init_vth!()`. Could we reinterpret it as phi is normalised by c\_ref^2/(e\*m), and T\_e is normalised by c\_ref^2/m (if we want to make ppar=n\*Tpar instead, and keep my change to `init_vth!()`)? (Normalised) T\_e is a settable parameter, so we can change it if we want to
https://github.com/mabarnes/moment_kinetics/blob/72ee32ed6691e14f553938e4a66c626955aa9218/src/moment_kinetics_input.jl#L91
https://github.com/mabarnes/moment_kinetics/blob/72ee32ed6691e14f553938e4a66c626955aa9218/src/em_fields.jl#L103

Could we also say that c\_ref=sqrt(T\_ref/m) for some T\_ref, without changing any of the code and normalise all temperatures by T\_ref and velocities by c\_ref? The only difference (?) would be that the values in SI that correspond to a given normalised value change for temperatures and velocities:
* T[SI, new norm] = 1/2 \* T[SI, old norm]
* v[SI, new norm] = 1/sqrt(2) \* v[SI, old norm]

It's looking to me like this choice doesn't matter (except in interpreting the input/output values in SI) until we add e.g. collision frequencies?

I'd prefer to have p=n\*T in normalised units. My first choice would be c\_ref=sqrt(T\_ref/m) because it means we can work out what the normalisations are by dimensional analysis without having to think about factors of 2. My second preference would be to leave c\_ref=sqrt(2\*T\_ref/m) but normalise all temperatures by m\*c\_ref^2 and phi by c\_ref^2/(m\*e) (i.e. keep the conventions as they are but with my update to `init_vth!()`). The current choice (with the original `init_vth!()` code) is confusing to me.

mabarnes:
I need to go through it more carefully, but my unease is around the following:

In the drift kinetic equation for ions, the term proportional to dphi/dz has a factor of 1/2 in it.  This is because we have set phi_norm = e * phi / T_r and assumed that T_r = 1/2 * m_r * c_r^2.  If we move to having T_r = m_r * c_r^2, then my concern is that the factor of 1/2 that is currently in the code would be incorrect.  That's fine, as we could just change it, but it's not clear to me that we could just re-interpret the definition of the reference temperature without modifying the equations somewhere.

I'm coming across all of this because with the electrons as an evolved species, T_e is no longer just a fixed input variable.  My code at the moment has kept things consistent with the init_vth! code that was there; i.e., with p_norm = 1/2 * n_norm * T_norm.  I agree that this is irritating but will probably leave it at least until this June report is finished.

mabarnes:
p.s. it seems to me that there are two clean options: define thermal speeds everywhere as sqrt(T/m) and normalise pressure, temperature and parallel heat flux without factors of 2 anywhere, or keep vth = sqrt(2T/m) and change the pressure and parallel heat flux normalisations appropriately.  I think both ways would get rid of random factors of 2 all over the place.

johnomotani:
@mabarnes I forgot about the 1/2 in the dphi/dz term. I agree with what you said. The original `init_vth!()` is consistent, and we are using p=1/2\*n\*T in normalised units (at the moment).

I'll take out the `init_vth!()` change.

> p.s. it seems to me that there are two clean options: define thermal speeds everywhere as sqrt(T/m) and normalise pressure, temperature and parallel heat flux without factors of 2 anywhere, or keep vth = sqrt(2T/m) and change the pressure and parallel heat flux normalisations appropriately. I think both ways would get rid of random factors of 2 all over the place.

Agree we should do one of these. Definitely deserves a PR of its own though, and can wait until after your report.

johnomotani:
I'll also change the parallel temperature plot to use Tpar=2*ppar/n, which should be consistent with what we are currently doing.

PR #101, opened 2023-08-01
Fix upar initialisation for manufactured solutions

johnomotani:
`init_upar!()` has been updated to actually calculate `upar`, so we should not divide by density after returning from the function.


mrhardman:
I am very happy with this approach. I started doing the same thing in my own development branch with the extra flourish that the lowest level calculation was a function that can be called from a naive Julia script without MPI for testing purposes. You can see what I mean here 
https://github.com/mabarnes/moment_kinetics/blob/042f122db459cd8ee483fcf1bc2db5daacaf27ff/src/velocity_moments.jl#L213C1-L233C4

I also put in the masses so that we could generalise later to the multispecies case without worrying again about normalisations. https://github.com/mabarnes/moment_kinetics/commit/5092c10b906225c40577a7097714085065968bee

PR #100, opened 2023-06-27
Automate setup on known HPC clusters

johnomotani:
Provide scripts that set up a working configuration for `moment_kinetics` on known HPC clusters (currently ARCHER2 and Marconi).

A wrapper script `machines/machine_setup.sh` prompts the user for some options (most of which have sensible defaults) and runs the complete setup: downloading Julia (if requested); setting up `JULIA_DEPOT_PATH` (necessary on ARCHER2 as the home directory is not accessible from the compute nodes); setting up HDF5 and MPI; and (if desired) launching a job to precompile, creating `moment_kinetics.so`.

The sub-steps run by `machine_setup.sh` can be run individually if that is useful, they are described in an [Advanced usage section in the docs](https://mabarnes.github.io/moment_kinetics/previews/PR100/machine_setup/#Advanced-usage).

Also includes a script `submit_run.sh` that submits a simulation and linked post-processing job, using template submission scripts. Run parameters (number of nodes, time limit, etc.) have defaults stored in `LocalPreferences.toml` (these can be set when running the setup, and changed by editing `LocalPreferences.toml`), but can be altered with command-line options to the script. The options are only single-letter flags, because I found the function to parse those with Bash and long option names seemed a lot more complicated - but they are documented in the help printed by `submit_run.sh -h`.

I made this PR because I got parallel I/O working on ARCHER2 and wanted to document the process. We've found having some cluster-specific setup scripts useful with BOUT++, so I thought they would be convenient for `moment_kinetics`.
Once I'd got ARCHER2 working I had another go at getting parallel I/O to work on Marconi, and managed it. Not sure what I did differently this time than the last time I tried and failed! Requires compiling HDF5 from the source code - a script is included that takes care of that as part of the automated setup.

The scripts are mostly a combination of bash and Julia, to avoid needing to install or setup any packages in order to use them (other than the Julia that is being set up by the script itself). The one exception is a Python script that is used to get the link to download Julia binaries. The best way to download the latest stable version of Julia (or any other version, there's a prompt asking if you want to pick one) involved parsing a json file that's provided by the Julia organisation listing some info for all the released versions. Python seemed like the most portable straightforward way to do that parsing. The Python script has no dependencies outside the core Python install though, so any Python>=3.6 should be enough to run it.

PR #102, opened 2023-08-03
Fixes for plot_charged_pdf_2D_at_wall()

johnomotani:
Errors were introduced in 1d-2d merge.


mrhardman:
The only improvement suggestion would be to avoid loading all time data at once, since the plots are just for the last time step. That doesn't seem to be an option considered in the load_data.jl function used here.

johnomotani:
> could you also please merge this into #100 so that I can continue testing that PR?

:+1: since it's more use having these changes in #100 than in master at the moment, I'm going to try to just merge this into there...

johnomotani:
> The only improvement suggestion would be to avoid loading all time data at once, since the plots are just for the last time step. That doesn't seem to be an option considered in the load_data.jl function used here.

Got distracted by something else and didn't see this comment. That is a good point, and the loading should be able to handle that. I'll make a PR (into #100 for convenience) because it should be simple.

PR #103, opened 2023-08-03
In plot_charged_pdf_2D_at_wall(), only load data for last time point

johnomotani:
...this is all that is needed for the plots.


johnomotani:
@mrhardman - I haven't tested this because I don't have convenient data to hand. Hopefully it's a small enough change that it'll work first time...

PR #104, opened 2023-08-09
Skip distributed MPI setup in mk_input() except when running simulation

johnomotani:
When `mk_input()` is called from `setup_moment_kinetics()`, it needs to set up distributed memory MPI, but otherwise (e.g. in post-processing) it is just used to recreate the inputs used for a run, so does not need to set up distributed MPI.

### DIFF
--- a/src/moment_kinetics.jl
+++ b/src/moment_kinetics.jl
@@ -312,7 +312,7 @@ function setup_moment_kinetics(input_dict::Dict; restart_prefix_iblock=nothing,
     # Set up MPI
     initialize_comms!()
 
-    input = mk_input(input_dict; save_inputs_to_txt=true)
+    input = mk_input(input_dict; save_inputs_to_txt=true, ignore_MPI=false)
     # obtain input options from moment_kinetics_input.jl
     # and check input to catch errors
     io_input, evolve_moments,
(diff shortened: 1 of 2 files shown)


mrhardman:
I confirm that this bugfix allows the post processing script to run in the distributed MPI case that I am considering.

mrhardman:
On this branch (on so presumably still in master) I get the following error when I try to use the Chodura condition diagnostic

post-processing runs/wall-bc_cheb_2D_sinpi_hfuel/ Fri 11 Aug 16:06:25 BST 2023
  Activating project at `//moment_kinetics_2`
[ Info: Saved animation to //moment_kinetics_2/runs/wall-bc_cheb_2D_sinpi_hfuel/wall-bc_cheb_2D_sinpi_hfuel_phi_vs_r.gif
ERROR: LoadError: type Tuple has no field T_e
Stacktrace:
 [1] getproperty(x::Tuple{moment_kinetics.input_structs.species_composition}, f::Symbol)
   @ Base ./Base.jl:38
 [2] analyze_and_plot_data(prefix::String; run_index::Nothing)
   @ moment_kinetics.post_processing //moment_kinetics_2/src/post_processing.jl:764
 [3] analyze_and_plot_data(prefix::String)
   @ moment_kinetics.post_processing //moment_kinetics_2/src/post_processing.jl:383
 [4] top-level scope
   @ //moment_kinetics_2/run_post_processing.jl:7
in expression starting at //moment_kinetics_2/run_post_processing.jl:1

PR #105, opened 2023-08-10
New post-processing module using Makie.jl

johnomotani:
I've written a new post-processing module. It aims to:
* Read post-processing options from a TOML file, so that the options can be changed without having to restart Julia, or precompile again.
* Use Makie.jl as the plotting backend. Makie.jl is pure Julia (so avoids the Python or other dependencies of Plots.jl) and because it's not a wrapper of a wrapper of a huge Python library the interface is nicer. Makie seems to behave much better in terms of not making axis labels run over the edge of the figure and things like that. You might hope that Makie would also be faster than Plots, but at least when we make animations, almost all of the time seems to be in the backend C library (FFMPEG, etc.), so in the very limited comparison I did there was little difference (Plots.jl was actually marginally faster).
* Provide more of an interactive interface. This part is still a bit work-in-progress as I haven't tried it out much (or at all) yet, but you can do something like
    ```julia
    julia> using moment_kinetics.makie_post_processing
    julia> run_info = get_run_info("runs/wall-bc_cheb")
    julia> setup_makie_post_processing_input!(run_info_moments=run_info)
    julia> plot_vs_r_z(run_info, phi)
    julia> plot_vs_r_z(run_info, phi, ir=42) # decided we want a not-default r-index for the plot
    ```
    It should work that plots pop up, and if you switch the backend from CairoMakie (which is the one for making pdfs) to GLMakie (the interative one) they should be interactive, but I haven't tried this yet (need to check if the functions return the right thing...).

~~This PR is a draft because documentation is definitely needed, and I'll probably rebase the branch when I add that in, rather than add more commits (so probably better not to merge into any branch you want to keep until I remove the 'Draft' status).~~
Docs now available here https://mabarnes.github.io/moment_kinetics/previews/PR105/makie_post_processing/

Almost all the functionality of the existing `post_processing` module is (I think!) reproduced, plus a load of extra plot/animation types (as I was able to programatically generate functions for all combinations of one or two dimensions). To find out what the available options are, run `generate_example_input_file()` to generate an example file with all the options present (with their default values), but commented out.

Because Plots.jl and CairoMakie.jl are both pretty heavy dependencies (they take a longish time to precompile), I think if we're going to merge this it might be a good idea to remove the existing `post_processing` module so we can drop `Plots.jl` completely. At least eventually (in the not-too-distant future) I think that's the way to go, although we could keep them side by side for a while if people want that.
[We could also replace Makie with Plots in this new module (there are not too many places where Makie functions are called explicitly, it's mostly wrapped away in functions like `plot_1d()`), or we could re-write `post_processing` to use Makie. I'm biased, but I'm quite fond of both Makie and my new `makie_post_processing` module, so I'm not voting for either re-writing option!]

Todo:
* [x] documentation!
* [ ] make plots of distribution function from moment-kinetic simulations on the unnormalised grid (this is the one piece of functionality still missing).
* [x] possibly refactor `plots_for_dfn_variable()` to loop over dimension combinations rather than having them all typed out explicitly.
* [ ] Remove `post_processing.jl`, remove `Plots.jl` dependency?


mrhardman:
This all sounds positive. I look forward to trying it out.

My main comment regarding future effort is that the post processing should module should be human-readable enough so that custom plots or new diagnostics can be added by most users. For that reason I would be keen for at least some of the lowest level plotting routines to be written out explicitly so that there are templates in the new framework which can be easily modified. If a user wanted to modify slightly one of the plots that exists inside the loop over all dimensions, it might be useful to make it obvious how that could be done.

PR #119, opened 2023-09-04
Precompile fixes and improvements

johnomotani:
* There was a bug in `util/precompile_makie_plots.jl` which meant that precompiling makie_post_processing would fail.
* Add an option in `submit-run.sh` and `submit-restart.sh` to just create submission scripts, not actually submit the jobs. This is useful if you want to edit the submission scripts by hand for some reason.
* Uses `run_makie_post_processing.jl` in the cluster post-processing job. ~~(To change this locally, edit `machines/archer/jobscript-postprocess.template`, etc.)~~ To use the original `post_processing.jl` instead, pass the flag `-o` to the `submit-run.sh` or `submit-restart.sh` scripts.
* Adds the `-O3` flag when compiling the sysimage for post-processing. This should not have any negative consequences (unlike `--check-bounds=no), and might speed things up.
* To avoid timeouts when running sysimage-compiling jobs:
    * Separates the sysimage-compiling for the simulation code and the post-processing into separate jobs (they are different sysimages because simulations use optimized flags, while post-processing should not use `--bounds-check=no`).
    * Adds an option to make a sysimage with just the precompilation step done, without an test simulations being run to compile functions used in the code. This should be faster (in case the sysimage-compile job is timing out) at the cost of increasing compile time at the beginning of each simulation.

### DIFF
--- a/machines/archer/jobscript-postprocess-plotsjl.template
+++ b/machines/archer/jobscript-postprocess-plotsjl.template
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+#SBATCH --mem=POSTPROCMEMORY
+#SBATCH --time=POSTPROCTIME
+#SBATCH --account=ACCOUNT
+#SBATCH --partition=serial
+#SBATCH --qos=serial
+#SBATCH --output=RUNDIRslurm-post-%j.out
+
+set -e
+
+cd $SLURM_SUBMIT_DIR
+
+# Get setup for Julia
+source julia.env
+
+# Workaround as cpus-per-task no longer inherited by srun from sbatch.
+# See https://docs.archer2.ac.uk/faq/upgrade-2023/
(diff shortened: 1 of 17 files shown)


johnomotani:
The precompile runs at least complete for me on this branch on ARCHER2. @mrhardman does this fix your issues?

johnomotani:
Precompiling and building sysimage (.so) files is still a work in progress, but no more ideas to try right now, so going to merge this. Note one thing that may help is to use the pre-release Julia version, julia-1.10.0-beta2, which parallelises the sysimage creation and can speed it up a lot (if trying this on a cluster, probably need to increase the `--ntasks` setting in the submission script).

One final tweak to this PR (mentioned in the description at the top now) is that I've added a flag `-o` for `submit-run.sh`/`submit-restart.sh` to use the original `post_processing.jl` instead of `makie_post_processing.jl`, if you prefer that.

PR #123, opened 2023-09-11
Unequally spaced element boundaries for grid optimisations

### LINKED ISSUE #122: Resolving the phi and Ez near z = z_wall
We typically observe a diverging Ez = - d phi / dz as z -> z_wall. Analytical considerations lead us to expect phi ~ sqrt(z - z_wall) as z -> z_wall. Using a finite-element grid with uniformly spaced element boundaries is potentially a wasteful use of resolution when we require many points near the wall, and few in the centre of the domain.

An optimisation of the code which permits flexibility is to change the grid specification to allow non-uniformly spaced element boundaries. 

Branch and PR where this is carried out will shortly be pushed.

mrhardman:
This pull request introduces flexibility into the specification of the element boundaries, naturally allowing packing of points near the wall boundaries for the purpose of capturing the square-root shape of the electric potential near the sheath entrance. The code modifications are minor -- in the specification of the grids, weights, and derivative routines there were variables called "scale_factor" and "shift". Using Gauss-Chebyshev-Lobatto points in the spectral element scheme, we can write these scale and shift factors in terms of differences and averages of the element boundary points. All that is required is to specify the desired locations of the element boundaries at the initial time of the simulation. Previously, element boundaries were assumed to be uniformly spaced.

Access the new functionality by specifying

`z_element_spacing_option = "sqrt" '

The default value is

`z_element_spacing_option = "uniform" '.

To see how this optimisation performs consider the output from the file "examples/wall-bc/wall-bc_cheb.toml", using the new element spacing option. Note that the electric field and potential are both very smooth near the sheath entrance. Here z_ngrid = 9 and z_nelement = 32.

[wall-bc_cheb_phi(r0,z)_vs_z.pdf](https://github.com/mabarnes/moment_kinetics/files/12577788/wall-bc_cheb_phi.r0.z._vs_z.pdf)
[wall-bc_cheb_Ez(r0,z)_vs_z.pdf](https://github.com/mabarnes/moment_kinetics/files/12577789/wall-bc_cheb_Ez.r0.z._vs_z.pdf)

We can obtain similar forms for the potential and electric field even down to low resolutions. For z_ngrid = 5 and z_nelement = 8 we can obtain the following plots, keeping all other input parameters fixed. Whilst the curves are not smooth, the resolution is more optimally used so that no large fluctuations are observed where the electric field diverges.

[wall-bc_cheb_lowres_Ez(r0,z)_vs_z.pdf](https://github.com/mabarnes/moment_kinetics/files/12577792/wall-bc_cheb_lowres_Ez.r0.z._vs_z.pdf)
[wall-bc_cheb_lowres_phi(r0,z)_vs_z.pdf](https://github.com/mabarnes/moment_kinetics/files/12577793/wall-bc_cheb_lowres_phi.r0.z._vs_z.pdf)

Closes #122

### DIFF
--- a/src/chebyshev.jl
+++ b/src/chebyshev.jl
@@ -41,6 +41,22 @@ end
 
 """
 initialize chebyshev grid scaled to interval [-box_length/2, box_length/2]
+we no longer pass the box_length to this function, but instead pass precomputed
+arrays element_scale and element_shift that are needed to compute the grid.
+
+ngrid -- number of points per element (including boundary points)
+nelement_local -- number of elements in the local (distributed memory MPI) grid
+n -- total number of points in the local grid (excluding duplicate points)
+element_scale -- the scale factor in the transform from the coordinates 
+                 where the element limits are -1, 1 to the coordinate where
+                 the limits are Aj = coord.grid[imin[j]-1] and Bj = coord.grid[imax[j]]
+                 element_scale = 0.5*(Bj - Aj)
+element_shift -- the centre of the element in the extended grid coordinate
+                 element_shift = 0.5*(Aj + Bj)
+imin -- the array of minimum indices of each element on the extended grid.
+        By convention, the duplicated points are not included, so for element index j > 1
+        the lower boundary point is actually imin[j] - 1
+imax -- the array of maximum indices of each element on the extended grid.
 """
 function scaled_chebyshev_grid(ngrid, nelement_local, n,
 			element_scale, element_shift, imin, imax)
@@ -55,14 +71,12 @@ function scaled_chebyshev_grid(ngrid, nelement_local, n,
     # setup the scale factor by which the Chebyshev grid on [-1,1]
     # is to be multiplied to account for the full domain [-L/2,L/2]
     # and the splitting into nelement elements with ngrid grid points
-    #scale_factor = 0.5*box_length/float(nelement_global)
+    
     # account for the fact that the minimum index needed for the chebyshev_grid
     # within each element changes from 1 to 2 in going from the first element
     # to the remaining elements
     k = 1
     @inbounds for j ∈ 1:nelement_local
-        #wgts[imin[j]:imax[j]] .= sqrt.(1.0 .- reverse(chebyshev_grid)[k:ngrid].^2) * scale_factor
-        # amount by which to shift the centre of this element from zero
         scale_factor = element_scale[j]
         shift = element_shift[j]
         # reverse the order of the original chebyshev_grid (ran from [1,-1])
@@ -88,11 +102,9 @@ function elementwise_derivative!(coord, ff, chebyshev::chebyshev_info)
     # check array bounds
     @boundscheck nelement == size(chebyshev.f,2) || throw(BoundsError(chebyshev.f))
     @boundscheck nelement == size(df,2) && coord.ngrid == size(df,1) || throw(BoundsError(df))
-    # note that one must multiply by 2*nelement/L to get derivative
-    # in scaled coordinate
-    #scale_factor = 2.0*float(coord.nelement_global)/coord.L
-	# scale factor is (length of a single element/2)^{-1}
-	
+    # note that one must multiply by a coordinate transform factor 1/element_scale[j] 
+    # for each element j to get derivative on the extended grid
+    
     # variable k will be used to avoid double counting of overlapping point
     # at element boundaries (see below for further explanation)
     k = 0
@@ -335,13 +347,9 @@ function clenshaw_curtis_weights(ngrid, nelement_local, n, imin, imax, element_s
         # scale to account for modified domain (not [-1,1])
         wgts[1:ngrid] = w*element_scale[1]
         if nelement_local > 1
-            # account for double-counting of points at inner element boundaries
-            #wgts[ngrid] *= 2
             for j ∈ 2:nelement_local
                 wgts[imin[j]-1:imax[j]] .+= w*element_scale[j]
             end
-            # remove double-counting of outer element boundary for last element
-            #wgts[n] *= 0.5
         end
     end
     return wgts
--- a/src/chebyshev.jl
+++ b/src/chebyshev.jl
@@ -63,8 +63,6 @@ function scaled_chebyshev_grid(ngrid, nelement_local, n,
     @inbounds for j ∈ 1:nelement_local
         #wgts[imin[j]:imax[j]] .= sqrt.(1.0 .- reverse(chebyshev_grid)[k:ngrid].^2) * scale_factor
         # amount by which to shift the centre of this element from zero
-        #iel_global = j + irank*nelement_local
-        #shift = box_length*((float(iel_global)-0.5)/float(nelement_global) - 0.5)
         scale_factor = element_scale[j]
         shift = element_shift[j]
         # reverse the order of the original chebyshev_grid (ran from [1,-1])
--- a/src/chebyshev.jl
+++ b/src/chebyshev.jl
@@ -42,8 +42,8 @@ end
 """
 initialize chebyshev grid scaled to interval [-box_length/2, box_length/2]
 """
-function scaled_chebyshev_grid(ngrid, nelement_global, nelement_local, n,
-			irank, box_length, imin, imax)
+function scaled_chebyshev_grid(ngrid, nelement_local, n,
+			element_scale, element_shift, imin, imax)
     # initialize chebyshev grid defined on [1,-1]
     # with n grid points chosen to facilitate
     # the fast Chebyshev transform (aka the discrete cosine transform)
@@ -55,7 +55,7 @@ function scaled_chebyshev_grid(ngrid, nelement_global, nelement_local, n,
     # setup the scale factor by which the Chebyshev grid on [-1,1]
     # is to be multiplied to account for the full domain [-L/2,L/2]
     # and the splitting into nelement elements with ngrid grid points
-    scale_factor = 0.5*box_length/float(nelement_global)
+    #scale_factor = 0.5*box_length/float(nelement_global)
     # account for the fact that the minimum index needed for the chebyshev_grid
     # within each element changes from 1 to 2 in going from the first element
     # to the remaining elements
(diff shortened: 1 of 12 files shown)


test/calculus_tests.jl
```diff
@@ -39,9 +39,11 @@ function runtests()
 				nrank_per_block = 0 # dummy value
 				irank = 0 # dummy value
 				comm = MPI.COMM_NULL # dummy value 
+                element_spacing_option = "uniform" # dummy value
```
johnomotani, line 42: Would be good to test "sqrt" spacing. I think it should pass this test? Instead of setting `element_spacing_option` here, replace line 17 above (Github won't let me put a comment that far away from changes) with
```julia
(discretization, element_spacing_option) ∈ (("finite_difference", "uniform"), ("chebyshev_pseudospectral", "uniform"), ("chebyshev_pseudospectral", "sqrt")),

File: test/calculus_tests.jl [hunk 2]
```diff
@@ -676,10 +690,12 @@ function runtests()
                 nelement_local = nelement
 				nrank_per_block = 0 # dummy value
 				irank = 0 # dummy value
-				comm = MPI.COMM_NULL # dummy value 
+				comm = MPI.COMM_NULL # dummy value
+                element_spacing_option = "uniform"
```
johnomotani, line 694: This test should work with a sqrt grid (the Chebyshev derivative tests just above should work too, but finding the expected errors is a faff, and those would probably change enough with the change of grid to make adding it there annoying), so remove this and add a
```julia
, element_spacing_option ∈ ("uniform", "sqrt")
```
to the test loop above.

test/calculus_tests.jl
```diff
@@ -726,10 +742,12 @@ function runtests()
...
```
johnomotani, line 746: test the sqrt spacing in this test, the same way as the previous comment.

test/interpolation_tests.jl
```diff
@@ -35,10 +35,12 @@ function runtests()
 			nelement_local = nelement
 			nrank_per_block = 0 # dummy value
 			irank = 0 # dummy value
-			comm = MPI.COMM_NULL # dummy value 
+			comm = MPI.COMM_NULL # dummy value
+            element_spacing_option = "uniform"
```
johnomotani, line 39: Could test sqrt spacing here. Change the `(discretization, rtol)` loop to `(discretization, element_spacing_option rtol)`...

test/wall_bc_tests.jl
```diff
@@ -191,11 +191,12 @@ function run_test(test_input, expected_phi, tolerance; args...)
         adv_input = advection_input("default", 1.0, 0.0, 0.0)
 		nrank_per_block = 0 # dummy value
 		irank = 0 # dummy value
-		comm = MPI.COMM_NULL # dummy value 
+		comm = MPI.COMM_NULL # dummy value
+        element_spacing_option = "uniform"
```
johnomotani, line 195: Would be good to add sqrt spacing to this test too. This one wouldn't check correctness, but acts as a regression check that we do not break the sqrt grid spacing by mistake.

Above, add something like
```julia
test_input_chebyshev_sqrt_grid = merge(test_input_chebyshev,
                                       Dict("run_name" => "chebyshev_pseudospectral_sqrt_grid",
                                            "z_element_spacing_option" => "sqrt")
```

And after the "Chebyshev" test below, add
```julia
        @testset "Chebyshev" begin
            run_test(test_input_chebyshev_sqrt_grid,
                     [0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
                     2.e-3)
```
when you run the test, the failure message should show what the computed values were, which you can then paste in instead of the 0.0's.

You'll also need to add `"z_element_spacing_option" => "uniform"` to the `test_input_finite_difference` Dict above.

test/wall_bc_tests.jl
```diff
@@ -191,11 +191,12 @@ function run_test(test_input, expected_phi, tolerance; args...)
...
         input = grid_input("coord", test_input["z_ngrid"], test_input["z_nelement"], 
 						   test_input["z_nelement"], nrank_per_block, irank, 1.0,
                            test_input["z_discretization"], "", test_input["z_bc"],
-                           adv_input, comm)
+                           adv_input, comm, element_spacing_option)
```
johnomotani, line 199:
```suggestion
                           adv_input, comm, test_input["z_element_spacing_option"])
```

johnomotani (commented):
Looks good to me! Thanks @mrhardman!

There are a few docstrings/comments to tidy up, and I think we can add a reasonable number of tests without too much work.

src/post_processing.jl
```diff
@@ -1054,7 +1055,7 @@ function analyze_and_plot_data(prefix...; run_index=nothing)
         composition, species, collisions,
         geometry, drive_input, num_diss_params, manufactured_solns_input = input
 
-    if !is_1D1V
+    if !is_1D1V || true
```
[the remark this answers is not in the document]
mrhardman: This is an accident -- but it does represent the fact that I do continue to use my 2D diagnostics for 1D1V simulations. I don't know how to extract individuals time slices from a gif, so there are certain plots in the 2D grouping that I continue to want printing out!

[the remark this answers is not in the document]
johnomotani: You won't be surprised that my answer to this is: it's not a problem in `makie_post_processing`! If we wanted to keep `post_processing`, we should really fix it so that plots that make sense in both 1D and 2D are always made, without duplicating code...

mrhardman:
@johnomotani  I have just pushed a commit which addresses your comments to the main code. I still have to address your suggestions regarding extending the tests. This appears to be slightly involved as the "sqrt" option appears not to pass the tests to the same tolerances as the "uniform" option. This should be expected in cases where the test function has variation in the centre of the test domain.

test/calculus_tests.jl
mrhardman: Making this change leads to tests which fail. I will look in to this as soon as I have bandwidth!

src/chebyshev.jl
```diff
@@ -55,24 +71,22 @@ function scaled_chebyshev_grid(ngrid, nelement_global, nelement_local, n,
     # setup the scale factor by which the Chebyshev grid on [-1,1]
     # is to be multiplied to account for the full domain [-L/2,L/2]
     # and the splitting into nelement elements with ngrid grid points
```
johnomotani, line 73: one last out-of-dat comment!
```suggestion
```

johnomotani:
About that test failure: the integration code looks right, but I'm wondering if there could be an inconsistency in how element boundaries are treated in derivatives and integrals.

It looks like (from the fact that the test passed on uniform grids, and I guess would pass with one element) on a single element we would have integral(derivative(f)) = f[end] - f[start]. Now what if we have two elements? The integration weights are calculated independently for each element and summed at the element boundary, so if the elements have different widths the total integration weight for the element boundary has unequal contributions from the elements on either side. However, in the derivative routine, the derivative at the element boundary is taken to be the mean of the derivatives in the elements on the left and right. To keep the property that numerically the integral of the derivative is exactly the change in f, I think the derivative at the element boundary would have to be some sort of weighted sum of the derivatives from either element, to match up with the unequal contributions to the integration weights.

The proposal in the previous paragraph would mean giving more (?) weight to the derivative from the wider element, so it's not obvious to me that this is what we want to do - presumably (at least for a non-special function whose steepness is not related to the widths of the elements) the smaller element would give a more accurate estimate of the derivative at the element boundary, so you might think we'd want to give the smaller element a larger weight in the derivative at the element boundary...

mrhardman:
If this is a matter of precision, and we can pass the tests with tolerances like 1.0e-8 or so, I would be in favour of marking this as an issue to be returned to, and reducing the tolerance in the specific tests using the "sqrt" option. I am not sure that spectral elements naturally satisfy the fundamental theorem of calculus in the general case. To the best of my understanding, no part of our code relies on this feature. To the best of my understanding, as we upwind the distribution function anyway, the fundamental theorem of calculus is broken in the kinetic equation, which means that density is not conserved anyway.

johnomotani:
Happy to reduce the tolerance for the sqrt spacing tests, and leave anything further as a future todo.

One motivation for 'fixing' this at some point: although at the moment we're hardly using the 'central' derivatives anywhere (mostly using the upwind ones wherever possible), we could use the 'fundamental theorem of calculus' property to conserve particles in moment kinetic modes. If we used a centred element-boundary in the continuity equation (I think @mabarnes did originally and I changed it to an upwinding version when trying to make things more stable) we could conserve particles because we could have numerically that
```math
\int_{-L/2}^{L/2}dz \frac{\partial(n u_\parallel)}{\partial z} = (nu_\parallel)|_{L/2} - (nu_\parallel)|_{-L/2}
```
As a bonus, the proposal for doing this would give a kind of partial upwinding - if element sizes get smaller towards the target, then at each element boundary we would be giving more weight to the derivative from the larger (upwind) element.

mrhardman:
@johnomotani Your comments should now be addressed, with at least some testing support for the "sqrt" `element_spacing_option'. Could you take a quick look and comment whether or not we can merge this into master?

test/calculus_tests.jl
```diff
@@ -39,9 +38,11 @@ function runtests()
 				nrank_per_block = 0 # dummy value
 				irank = 0 # dummy value
 				comm = MPI.COMM_NULL # dummy value 
+                #element_spacing_option = "uniform" # dummy value
```
johnomotani, line 41:
```suggestion
```

test/wall_bc_tests.jl
```diff
@@ -191,16 +218,20 @@ function run_test(test_input, expected_phi, tolerance; args...)
...
                            test_input["z_discretization"], "", test_input["z_bc"],
-                           adv_input, comm)
+                           adv_input, comm, test_input["z_element_spacing_option"])
         z, z_spectral = define_coordinate(input)
 
         # Cross comparison of all discretizations to same benchmark
-        phi_interp = interpolate_to_grid_z(cross_compare_points, phi[:, end], z, z_spectral)
-        @test isapprox(phi_interp, cross_compare_phi, rtol=tolerance, atol=1.e-15)
+        if test_input["z_element_spacing_option"] == "uniform" 
+            # only support this test for uniform element spacing
```
johnomotani, line 231:
```suggestion
            # Only support this test for uniform element spacing.
            # phi is better resolved by "sqrt" spacing grid, so disagrees with benchmark data from
            # simulation with uniform element spacing.
```

johnomotani (approved):
Looks good to me. Thanks @mrhardman! We can merge once the tests complete.

PR #127, opened 2023-09-11
Krook 'collisions'

johnomotani:
Add a Krook collision operator. Needed this to make 1D1V simulations look nicer for EFTC!

Also includes a script to print some dimensional parameters from a simulation (e.g. ion-ion collision frequency).

TODO:
* [x] test - at least a regression test, including the moment-kinetic versions, is needed


johnomotani:
@mrhardman I'm proposing to merge this PR, as this Krook collision operator supports moment-kinetic modes, and has a density/temperature dependent collision frequency. Are you OK with that? It will need some merging with your MMS tests when you make a PR for those.

mrhardman:
> @mrhardman I'm proposing to merge this PR, as this Krook collision operator supports moment-kinetic modes, and has a density/temperature dependent collision frequency. Are you OK with that? It will need some merging with your MMS tests when you make a PR for those

I would prefer the MMS test options to be included now, since it seems that you only have a regression test here. The MMS tests are quite orthogonal to the rest of my collision operator development in that branch. Can we discuss offline what the barriers to including the MMS tests are?

src/krook_collisions.jl
```diff
@@ -0,0 +1,134 @@
...
+
+"""
+Add collision operator
+
+Currently Krook collisions
+"""
+function krook_collisions!(pdf_out, fvec_in, moments, composition, collisions, vperp, vpa, dt)
+    begin_s_r_z_region()
+
+    if vperp.n > 1
```
mrhardman, line 45: Are 2V collisions implemented now? This warning suggests not.

src/moment_kinetics_input.jl
```diff
@@ -197,27 +92,26 @@ function mk_input(scan_input=Dict(); save_inputs_to_txt=false, ignore_MPI=true)
     composition.Er_constant = get(scan_input, "Er_constant", 0.0)
 
     # Get reference parameters for normalizations
-    reference_parameter_section = set_defaults_and_check_section!(
+    reference_parameter_section = copy(set_defaults_and_check_section!(
```
mrhardman, line 95: This is probably the wrong place to discuss this, but I don't understand why we need reference parameters in the code itself.

src/krook_collisions.jl
```diff
@@ -0,0 +1,134 @@
...
+        @loop_s_r_z is ir iz begin
+            n = fvec_in.density[iz,ir,is]
+            vth = moments.charged.vth[iz,ir,is]
+            T = vth^2
+            if vperp.n == 1
+                vth_prefactor = 1.0 / vth
+            else
+                vth_prefactor = 1.0 / vth^3
+            end
+            nu_ii = collisions.krook_collision_frequency_prefactor * n * T^(-1.5)
```
mrhardman, line 120: This choice of prefactor is probably motivated by the physical form of Fokker-Planck collision frequency, but it is not obvious to me that we would actually want this feature. Isn't it easier to understand the ad-hoc feature that we add if it doesn't have a nonlinear collision frequency?

src/krook_collisions.jl
johnomotani: Ah, put this in before I implemented them! Should just apply to the moment-kinetic cases now.

src/krook_collisions.jl
```diff
@@ -0,0 +1,134 @@
...
+"""
+Add collision operator
+
+Currently Krook collisions
+"""
+function krook_collisions!(pdf_out, fvec_in, moments, composition, collisions, vperp, vpa, dt)
+    begin_s_r_z_region()
+
+    if vperp.n > 1
+        error("Krook collisions not implemented for 2V case yet")
```
johnomotani, line 46:
```suggestion
    if vperp.n > 1 && (moments.evolve_density || moments.evolve_upar || moments.evolve_ppar)
        error("Krook collisions not implemented for 2V moment-kinetic cases yet")
```

mrhardman:
I am starting to merge in the appropriate features. The job may take some time as even the manufactured solutions module appears to have diverged from the initial state where https://github.com/mabarnes/moment_kinetics/commit/d5c366fafec352f64f4e84e328808c77dc0246a5 was committed to. If this merge is urgent, we should discuss.

src/krook_collisions.jl
johnomotani: I think it makes sense to have both options (e.g. Gkeyll does that https://gkeyll.readthedocs.io/en/latest/gkyl/App/Collisions/collisionModels.html). If you want to add an option for constant collision frequency I'm happy for that to be there. I wanted to have a density/temperature dependent collision frequency because in a setup where physically the collisionality changes significantly (e.g. due to a large temperature drop between midplane and target) I don't want to have unphysically strong collisions upstream, but want the collisions near the target to be strong enough to Maxwellian-ise the plasma.

src/moment_kinetics_input.jl
johnomotani: My reasoning is that we need the reference parameters to calculate the collision frequency, so for me it's easier to understand if it's explicit what the reference parameters are for a given simulation (i.e. given in the input file or set by defaults), and 'everything else' is forced to be consistent with them. At the moment it's 'just' (I guess?) rhostar and collision frequency at reference parameters that are actually being used, but at some point we might need/want for example atomic rates (ionization, charge exchange, etc.) that are tabulated functions of density and temperature.

Another motivation is that we might want a post processing routine that calculates things in SI units, like the heat flux to the target (in MW/m^2), or the power input into the simulation (in MW). Then we'd need to know what the reference parameters are, so it's simplest if they are explicitly specified.

What's the downside of having reference parameters specified?

src/krook_collisions.jl
johnomotani: Fixed now!

mrhardman:
> What's the downside of having reference parameters specified?

So long as it is clear that all evolved quantities are normalised, I will have to be happy. In HPC codes that I have worked with previously all small numbers are eliminated from the equations in favour of normalised parameters which enter the input file, and are computed externally to the HPC code. This also gave the option of making physics studies where unphysical parameter scans can be supported (changing the mass of particles, but not collision frequency, for example). In an ideal implementation we would use all reference parameters in a separate module and not propagate their use into the main code, where only normalised quantities are used. This might be what you are already doing -- in which case I cannot find fault.

johnomotani:
> So long as it is clear that all evolved quantities are normalised, I will have to be happy. In HPC codes that I have worked with previously all small numbers are eliminated from the equations in favour of normalised parameters which enter the input file, and are computed externally to the HPC code. This also gave the option of making physics studies where unphysical parameter scans can be supported (changing the mass of particles, but not collision frequency, for example). In an ideal implementation we would use all reference parameters in a separate module and not propagate their use into the main code, where only normalised quantities are used. This might be what you are already doing -- in which case I cannot find fault.

Agree with all of this. I would say that I think it's nicer if 'unphysical' setups have to be specified explicitly, with physically consistent things being the default. It does seem like a good idea to separate the 'reference parameters' into a separate module so that the separation is more obviously enforced - that they are defined, but only used in one or two places within the setup. I'll make a commit to tidy this up a bit...

mrhardman:
An initial attempt to port the modifications to the manufactured solutions module to permit the MMS testing of the Krook operator is here: https://github.com/mabarnes/moment_kinetics/commit/030fbe780dd93d3d0893b2ed8dae54314297440a

The O(1) errors in the test of u|| and p|| likely arise from conflicting normalisation conventions.

src/krook_collisions.jl
```diff
@@ -0,0 +1,134 @@
...
+    # Note: do not need 1/sqrt(pi) for the 'Maxwellian' term because the pdf is already
+    # normalized by sqrt(pi) (see velocity_moments.integrate_over_vspace).
+
+    if moments.evolve_ppar && moments.evolve_upar
+        # Compared to evolve_upar version, grid is already normalized by vth, and multiply
+        # through by vth, remembering pdf is already multiplied by vth
+        @loop_s_r_z is ir iz begin
+            n = fvec_in.density[iz,ir,is]
+            T = (moments.charged.vth[iz,ir,is])^2
+            nu_ii = collisions.krook_collision_frequency_prefactor * n * T^(-1.5)
```
johnomotani, line 58: Rename to make clear this is not the 'normalised temperature' as currently defined for the code.
```suggestion
            T_over_Tref = (moments.charged.vth[iz,ir,is])^2
            nu_ii = collisions.krook_collision_frequency_prefactor * n * T_over_Tref^(-1.5)
```

src/krook_collisions.jl
```diff
@@ -0,0 +1,134 @@
...
+            end
+        end
+    elseif moments.evolve_ppar
+        # Compared to full-f collision operater, multiply through by vth, remembering pdf
+        # is already multiplied by vth, and grid is already normalized by vth
+        @loop_s_r_z is ir iz begin
+            n = fvec_in.density[iz,ir,is]
+            vth = moments.charged.vth[iz,ir,is]
+            T = vth^2
+            nu_ii = collisions.krook_collision_frequency_prefactor * n * T^(-1.5)
```
johnomotani, line 72:
```suggestion
            T_over_Tref = vth^2
            nu_ii = collisions.krook_collision_frequency_prefactor * n * T_over_Tref^(-1.5)
```

src/krook_collisions.jl
```diff
@@ -0,0 +1,134 @@
...
+                           - (vperp.grid[ivperp]/vth)^2))
+            end
+        end
+    elseif moments.evolve_upar
+        # Compared to evolve_density version, grid is already shifted by upar
+        @loop_s_r_z is ir iz begin
+            n = fvec_in.density[iz,ir,is]
+            vth = moments.charged.vth[iz,ir,is]
+            T = vth^2
+            nu_ii = collisions.krook_collision_frequency_prefactor * n * T^(-1.5)
```
johnomotani, line 86:
```suggestion
            T_over_Tref = vth^2
            nu_ii = collisions.krook_collision_frequency_prefactor * n * T_over_Tref^(-1.5)
```

src/krook_collisions.jl
```diff
@@ -0,0 +1,134 @@
...
+            end
+        end
+    elseif moments.evolve_density
+        # Compared to full-f collision operater, divide through by density, remembering
+        # that pdf is already normalized by density
+        @loop_s_r_z is ir iz begin
+            n = fvec_in.density[iz,ir,is]
+            vth = moments.charged.vth[iz,ir,is]
+            T = vth^2
+            nu_ii = collisions.krook_collision_frequency_prefactor * n * T^(-1.5)
```
johnomotani, line 101:
```suggestion
            T_over_Tref = vth^2
            nu_ii = collisions.krook_collision_frequency_prefactor * n * T_over_Tref^(-1.5)
```

src/krook_collisions.jl
```diff
@@ -0,0 +1,134 @@
...
+        @loop_s_r_z is ir iz begin
+            n = fvec_in.density[iz,ir,is]
+            vth = moments.charged.vth[iz,ir,is]
+            T = vth^2
+            if vperp.n == 1
+                vth_prefactor = 1.0 / vth
+            else
+                vth_prefactor = 1.0 / vth^3
+            end
+            nu_ii = collisions.krook_collision_frequency_prefactor * n * T^(-1.5)
```
johnomotani, line 120:
```suggestion
            T_over_Tref = vth^2
            if vperp.n == 1
                vth_prefactor = 1.0 / vth
            else
                vth_prefactor = 1.0 / vth^3
            end
            nu_ii = collisions.krook_collision_frequency_prefactor * n * T_over_Tref^(-1.5)
```

src/krook_collisions.jl
```diff
@@ -0,0 +1,138 @@
...
+    else # default option, collisions.krook_collisions_option == "reference_parameters"
+        cfac = 1.0
+    end
+    if moments.evolve_ppar && moments.evolve_upar
+        # Compared to evolve_upar version, grid is already normalized by vth, and multiply
+        # through by vth, remembering pdf is already multiplied by vth
+        @loop_s_r_z is ir iz begin
+            n = fvec_in.density[iz,ir,is]
+            T = (moments.charged.vth[iz,ir,is])^2
+            nu_ii = collisions.krook_collision_frequency_prefactor * ( n * T^(-1.5) * cfac + (1.0 - cfac))
```
johnomotani, line 62:
```suggestion
            T_over_Tref = (moments.charged.vth[iz,ir,is])^2
            nu_ii = collisions.krook_collision_frequency_prefactor * ( n * T_over_Tref^(-1.5) * cfac + (1.0 - cfac))
```

src/krook_collisions.jl
```diff
@@ -0,0 +1,138 @@
...
+            end
+        end
+    elseif moments.evolve_ppar
+        # Compared to full-f collision operater, multiply through by vth, remembering pdf
+        # is already multiplied by vth, and grid is already normalized by vth
+        @loop_s_r_z is ir iz begin
+            n = fvec_in.density[iz,ir,is]
+            vth = moments.charged.vth[iz,ir,is]
+            T = vth^2
+            nu_ii = collisions.krook_collision_frequency_prefactor * ( n * T^(-1.5) * cfac + (1.0 - cfac))
```
johnomotani, line 76:
```suggestion
            T_over_Tref = vth^2
            nu_ii = collisions.krook_collision_frequency_prefactor * ( n * T_over_Tref^(-1.5) * cfac + (1.0 - cfac))
```

src/krook_collisions.jl
```diff
@@ -0,0 +1,138 @@
...
+                           - (vperp.grid[ivperp]/vth)^2))
+            end
+        end
+    elseif moments.evolve_upar
+        # Compared to evolve_density version, grid is already shifted by upar
+        @loop_s_r_z is ir iz begin
+            n = fvec_in.density[iz,ir,is]
+            vth = moments.charged.vth[iz,ir,is]
+            T = vth^2
+            nu_ii = collisions.krook_collision_frequency_prefactor * ( n * T^(-1.5) * cfac + (1.0 - cfac))
```
johnomotani, line 90:
```suggestion
            T_over_Tref = vth^2
            nu_ii = collisions.krook_collision_frequency_prefactor * ( n * T_over_Tref^(-1.5) * cfac + (1.0 - cfac))
```

src/krook_collisions.jl
```diff
@@ -0,0 +1,138 @@
...
+            end
+        end
+    elseif moments.evolve_density
+        # Compared to full-f collision operater, divide through by density, remembering
+        # that pdf is already normalized by density
+        @loop_s_r_z is ir iz begin
+            n = fvec_in.density[iz,ir,is]
+            vth = moments.charged.vth[iz,ir,is]
+            T = vth^2
+            nu_ii = collisions.krook_collision_frequency_prefactor * ( n * T^(-1.5) * cfac + (1.0 - cfac))
```
johnomotani, line 105:
```suggestion
            T_over_Tref = vth^2
            nu_ii = collisions.krook_collision_frequency_prefactor * ( n * T_over_Tref^(-1.5) * cfac + (1.0 - cfac))
```

src/krook_collisions.jl
```diff
@@ -0,0 +1,138 @@
...
+        @loop_s_r_z is ir iz begin
+            n = fvec_in.density[iz,ir,is]
+            vth = moments.charged.vth[iz,ir,is]
+            T = vth^2
+            if vperp.n == 1
+                vth_prefactor = 1.0 / vth
+            else
+                vth_prefactor = 1.0 / vth^3
+            end
+            nu_ii = collisions.krook_collision_frequency_prefactor * ( n * T^(-1.5) * cfac + (1.0 - cfac))
```
johnomotani, line 124:
```suggestion
            T_over_Tref = vth^2
            if vperp.n == 1
                vth_prefactor = 1.0 / vth
            else
                vth_prefactor = 1.0 / vth^3
            end
            nu_ii = collisions.krook_collision_frequency_prefactor * ( n * T_over_Tref^(-1.5) * cfac + (1.0 - cfac))
```

johnomotani (commented):
Make clearer that `T_over_Tref` is not the normalised temperature as defined in the code.

mrhardman:
> Make clearer that `T_over_Tref` is not the normalised temperature as defined in the code.

Why not avoid using T_over_Tref entirely and put 

`n * vth^(-3)`

in the collision operator factor?

mrhardman (commented):
Do we have a diagnostic output for what the actual collision frequency used is for this option? Could be helpful for understanding whether or not the inputs give O(1) coefficients in the MMS test.

johnomotani:
Implementing collision operator in terms of vth - I like that idea. I guess however we end up defining `cref`, we'd choose to have `vth(norm) = vth(unnorm)/cref`, so the implementation in terms of vth would keep working for any choice.

Diagnostic output for collision frequency - that would be useful, and should be an easy thing to implement.

johnomotani:
...both upgrades now made.

After adding a function that calculates collision frequency (and a little bit of refactoring of `makie_post_processing`), adding plots of the collision frequency in `makie_post_processing` only took 9 lines :sunglasses:
https://github.com/mabarnes/moment_kinetics/blob/c7735e160c3082053fc1d2ba3b14ea5e201d6ec3/src/makie_post_processing.jl#L73
https://github.com/mabarnes/moment_kinetics/blob/c7735e160c3082053fc1d2ba3b14ea5e201d6ec3/src/makie_post_processing.jl#L1210-L1213
https://github.com/mabarnes/moment_kinetics/blob/c7735e160c3082053fc1d2ba3b14ea5e201d6ec3/src/makie_post_processing.jl#L1333-L1336

johnomotani:
PS the spatially-varying collision frequency is O(1), at least in the 1V test
1D-wall_MMS_new_nel_r_1_z_16_vpa_16_vperp_1_krook_collision_frequency_spec1_vs_z

PR #128, opened 2023-09-11
Allow interpolation between grids when restarting

johnomotani:
Can be used to change resolution, and/or to switch between different moment-kinetic modes, when restarting.
Edit: Can also switch (in a basic way, see the docs) from 1D to 2D, or from 1V to 2V/3V.

This PR also includes some refactoring of the types used to dispatch to different discretization methods for differentiation and interpolation - this is a tidy-up and completion of the refactoring started in #57. Now there is an 'abstract type' `discretization_info`, defined in `moment_kinetics_structs`, which allows `chebyshev_info` (now a sub-type of `discretization_info`) to be moved back into `chebyshev.jl`, and replaces the use of `Bool` to indicate finite-difference methods with a `finite_difference_info` type (also a subtype of `discretization_info`).

TODO:
* [x] tests!
* [x] documentation


johnomotani:
[Copying this to the PR page because that will probably be easier to find in future!]

> Does this commit allow for one to run a 1D1V input file to steady state (with r_ngrid = 1 and r_nelement = 1), and then restart it with r_ngrid > 1 and r_nelement > 1. This would be a great timesaver! 2D1V simulations might be more reliable if we start with a better initial condition -- which we might not be able to guess or write in terms of closed-form functions.

I haven't considered going from 1D to 2D. It should be possible, but will need a bit of work. The interpolation routines were written with velocity space interpolation in mind, so when they have to go past the end of the input grid they fill in values with an exp(-v^2) decay from the last value on the grid. I guess we'd want to add an option to control the extrapolation, which could be set for the r-dimension to something like 'use the last value on the grid'. I'm also not sure what exactly happens with a `coordinate` struct for a dimension with just one point. I remember that we do some sort of special-case for it, but that might break interpolation somehow. On the other hand, maybe we could take advantage of that to do a special-case interpolation method where if the dimension had only one point, then the 'interpolation' is just to copy the single value to the whole output array.

mrhardman:
> On the other hand, maybe we could take advantage of that to do a special-case interpolation method where if the dimension had only one point, then the 'interpolation' is just to copy the single value to the whole output array

This is what I had in mind. The special exception that you mention is in coordinates.jl -- grids with coord.n = 1 have their wgt[1] set to a constant and grid[1] set to 0.0. If we just plan to copy the data in the "radial dimension" from r.n = 1 to all the other indices, I would have thought we would have a fairly straight forward operation to do.

johnomotani:
This branch now includes tests (https://github.com/mabarnes/moment_kinetics/blob/f13e3e03da956e775236d1a6d458eacb31dfcc9b/test/restart_interpolation_tests.jl) and handles interpolation from 1D to 2D and from 1V to 2V or 3V (48dcc80).

Should be ready to merge now, I hope.

PR #130, opened 2023-09-11
Add bc that sets zero only at the incoming boundary

johnomotani:
...even when diffusion is present. This is in contrast to "zero" which sets zero at the incoming boundary when no diffusion is present, but zero at both boundaries when some diffusion is present. This may be useful when numerical diffusion is present but is small compared to the advection speed near the boundary. Specify with, e.g. `vpa_bc = "incoming_zero"`.

Not sure whether this is really useful - I ended up not needing any numerical dissipation in my 1D1V runs for EFTC, so the existing `"zero"` bc was fine in the end. I guess we might as well add the option to force 'zero at the incoming boundary' in case it is useful in future.

### DIFF
--- a/docs/src/external_sources.md
+++ b/docs/src/external_sources.md
@@ -1,0 +1,6 @@
+`external_sources`
+==================
+
+```@autodocs
+Modules = [moment_kinetics.external_sources]
+```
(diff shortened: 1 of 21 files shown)


johnomotani:
Not sure this is really worth adding. It's easy to re-implement if it is needed in future.

PR #131, opened 2023-09-11
Save the run-time of simulations to the output file

johnomotani:
Save some run-time information to the output files: time for setup (excluding file I/O), and cumulative run time (excluding setup) at each output step.


johnomotani:
I'd based on the previous PRs, but that was just because it was the last thing I'd added in the stuff I'd been working on. Can rebase onto master easily, then should be merge-able (I hope the test failures were caused by a different PR, so should go away then). Will do that now.

PR #132, opened 2023-09-11
If a NaN is found, stop the simulation

johnomotani:
When writing output, check for NaNs and Allreduce/Bcast the result so that the simulation can stop cleanly. This probably won't cost much, and prevents time/energy being wasted running simulations that are silently producing junk.

### DIFF
--- a/src/time_advance.jl
+++ b/src/time_advance.jl
@@ -8,8 +8,8 @@ export time_advance!
 using MPI
 using ..type_definitions: mk_float
 using ..array_allocation: allocate_float, allocate_shared_float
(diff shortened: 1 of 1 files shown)


johnomotani:
I'd based on the previous PRs, but that was just because it was the last thing I'd added in the stuff I'd been working on. Can rebase onto master easily, then should be merge-able (I hope the test failures were caused by a different PR, so should go away then). Will do that now.

johnomotani:
I think the CI tests just got confused when I changed the base branch. The 'push' tests pass fine, but the 'pull' tests are getting a bug from an old version of `save-time-for-run`. Going to merge and hope it's OK.

PR #135, opened 2023-09-19
Simplify restart functionality

johnomotani:
Move actual implementation of restart functionality inside setup_moment_kinetics!() - this also fixes a bug where `iblock_index[]` was not initialised when `get_backup_filename()` needed to use it.

This allows the `restart_moment_kinetics()` function to be removed. It is replaced by a couple of arguments to run_moment_kinetics.

The `restart_moment_kinetics.jl` script is removed. Now either pass the `--restart` flag to `run_moment_kinetics()` to restart from the latest file in the run directory, or pass the output file to restart from as the second positional argument.

### DIFF
--- a/src/moment_kinetics.jl
+++ b/src/moment_kinetics.jl
@@ -2,7 +2,7 @@
 """
 module moment_kinetics
 
-export run_moment_kinetics, restart_moment_kinetics
+export run_moment_kinetics
 
 using MPI
 
@@ -89,11 +89,13 @@ using .type_definitions: mk_int
 """
 main function that contains all of the content of the program
 """
-function run_moment_kinetics(to::TimerOutput, input_dict=Dict())
+function run_moment_kinetics(to::TimerOutput, input_dict=Dict(); restart=false,
+                             restart_time_index=-1)
     mk_state = nothing
     try
         # set up all the structs, etc. needed for a run
-        mk_state = setup_moment_kinetics(input_dict)
+        mk_state = setup_moment_kinetics(input_dict; restart=restart,
+                                         restart_time_index=restart_time_index)
 
         # solve the 1+1D kinetic equation to advance f in time by nstep time steps
         if run_type == performance_test
@@ -136,27 +138,38 @@ end
 """
 overload which takes a filename and loads input
 """
-function run_moment_kinetics(to::TimerOutput, input_filename::String)
-    return run_moment_kinetics(to, read_input_file(input_filename))
+function run_moment_kinetics(to::TimerOutput, input_filename::String; restart=false,
+                             restart_time_index=-1)
+    return run_moment_kinetics(to, read_input_file(input_filename); restart=restart,
+                               restart_time_index=restart_time_index)
 end
 
 """
 overload with no TimerOutput arguments
 """
-function run_moment_kinetics(input)
-    return run_moment_kinetics(TimerOutput(), input)
+function run_moment_kinetics(input; restart=false, restart_time_index=-1)
+    return run_moment_kinetics(TimerOutput(), input; restart=restart,
+                               restart_time_index=restart_time_index)
 end
 
 """
 overload which gets the input file name from command line arguments
 """
 function run_moment_kinetics()
-    inputfile = get_options()["inputfile"]
-    if inputfile == nothing
-        run_moment_kinetics(Dict())
-    else
-        run_moment_kinetics(inputfile)
+    options = get_options()
+    inputfile = options["inputfile"]
+    restart = options["restart"]
+    if options["restart_file"] !== nothing
+        restart = options["restartfile"]
     end
+    restart_time_index = options["restart_time_index"]
+    if inputfile === nothing
+        this_input = Dict()
+    else
+        this_input = inputfile
+    end
+    run_moment_kinetics(this_input; restart=restart,
+                        restart_time_index=restart_time_index)
 end
 
 """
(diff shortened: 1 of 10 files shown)


johnomotani:
Not tested this yet. @mrhardman does it fix the bug you were running into?

mrhardman:
Attempting to start a normal simulation, I find a bug

ERROR: LoadError: KeyError: key "restart_file" not found
Stacktrace:
 [1] getindex(h::Dict{String, Any}, key::String)
   @ Base ./dict.jl:481
 [2] run_moment_kinetics()
   @ moment_kinetics /src/moment_kinetics.jl:162
 [3] top-level scope
   @ /run_moment_kinetics.jl:7
in expression starting at run_moment_kinetics.jl:7

src/moment_kinetics.jl
```diff
@@ -136,27 +138,38 @@ end
...
 overload which gets the input file name from command line arguments
 """
 function run_moment_kinetics()
-    inputfile = get_options()["inputfile"]
-    if inputfile == nothing
-        run_moment_kinetics(Dict())
+    options = get_options()
+    inputfile = options["inputfile"]
+    restart = options["restart"]
+    if options["restart_file"] !== nothing
```
mrhardman, line 162: replace "restart_file" with "restartfile" to be consistent with src/command_line_options.jl line 18

mrhardman (changes requested):
Some small typos in moment_kinetics.jl need correcting as indicated in comments. The parallel_io = false options now seem to be working for me if I use only a single shared memory region. Otherwise, for multiple shared memory regions using the distribution memory for z, the simulation crashes before the time advance with a bounds error but no stacktrace. In my own branch, the commits to fix the small bugs here are 8de082d594b9f72c14241be1dfc0685780fb93f5. A comment is modified in ce1b1346281a1f8084dd2505ba6465064c67c6f1. It would be useful to add a full example command to clarify the instructions on using the restart functionality in the manual.

src/moment_kinetics.jl
```diff
@@ -416,10 +326,41 @@ function setup_moment_kinetics(input_dict::Dict; restart_prefix_iblock=nothing,
...
+            io_settings = get(input_dict, "output", Dict{String,Any}())
+            binary_format = get(io_settings, "binary_format", hdf5)
+            if binary_format === hdf5
+                ext = "h5"
+            elseif binary_format === netcdf
+                ext = "cdf"
+            else
+                error("Unrecognized binary_format '$binary_format'")
+            end
+            restart = glob(joinpath(output_dir, run_name * ".dfns*." * ext))[1]
```
mrhardman, line 342: This line appears to be giving BoundsError when the number of shared memory regions is greater that 1.

src/moment_kinetics.jl
```diff
@@ -416,10 +326,41 @@ function setup_moment_kinetics(input_dict::Dict; restart_prefix_iblock=nothing,
...
+        end
+
+        # Move the output file being restarted from to make sure it doesn't get
+        # overwritten.
+        dfns_filename, backup_dfns_filename, parallel_io, moments_filename,
+        backup_moments_filename, backup_prefix_iblock =
+            get_backup_filename(restart)
```
mrhardman, line 349: Would prefer not to use `restart' as a logical and as a filename here.

src/moment_kinetics.jl
mrhardman: Does this code make any allowances for multiple .h5 or .cdf files to be copied and read? I cannot see how it does here.

mrhardman: Trying 

println(joinpath(output_dir, run_name * ".dfns*." * ext))
println(glob(joinpath(output_dir, run_name * ".dfns*." * ext)))

Gets 

runs/wall-bc_cheb_single_ion/wall-bc_cheb_single_ion.dfns*.h5
String[]

It's pretty clear that this code is broken for some reason in this case. What should it be doing? What does the function `glob' do? I cannot find the definition.

johnomotani: `glob()` finds files matching a pattern (name comes from *nix shell, e.g. https://tldp.org/LDP/abs/html/globbingref.html). The bounds error usually means there is no file matching the pattern it is looking for.

Some of the error messages can be a bit unhelpful. I tried to improve things a bit in a few commits in #128. Those would need merging if we merge this PR first - maybe I should just cherry-pick them into this branch...

src/moment_kinetics.jl
johnomotani: Can agree it's maybe not the clearest to change the value of `restart` that was passed in as an argument. Will make a separate `restart_filename` variable that gets set appropriately.

johnomotani:
@mrhardman I think I've addressed all your comments now.

johnomotani:
PS @mrhardman I don't think you've pushed the commits you mentioned in one of your comments. If there was a comment that needed updating, would you please just push the change to this branch? Thanks!

mrhardman:
> PS @mrhardman I don't think you've pushed the commits you mentioned in one of your comments. If there was a comment that needed updating, would you please just push the change to this branch? Thanks!

Sure! I was trying to debug these features by merging your changes into my own PR (generalised-chodura-diagnostic), since the simulations that I wanted to restart use a feature that is only in that branch at the moment. Switching branches seems to sometimes confuse my Julia compiler, so I will leave comment here directly.

mrhardman:
My comment was just that this line https://github.com/mabarnes/moment_kinetics/blob/7ba85583f3fc11b8c40ffe8f8093e1fda6a9c48b/src/command_line_options.jl#L19 should mention HDF5.

mrhardman:
I am happy now that the restart feature is working with parallel_io = false. My tests were carried out after merging this branch with https://github.com/mabarnes/moment_kinetics/pull/110. I have now pushed my commits there -- I didn't have to make any further modifications.

mrhardman (commented):
Great work, the restarting feature now works for parallel_io = false, and the code is easier to follow :  )

PR #147, opened 2023-11-17
Inputs and scripts for @johnomotani's poster at EFTC 2023

johnomotani:
Add a subdirectory `publication_inputs/2023_EFTC_jto-poster` which contains the inputs and scripts needed to create the figures for @johnomotani's poster at EFTC 2023. Includes a README.md with instructions how to recreate the runs and figures.


johnomotani:
Question @mabarnes @mrhardman - is this something that we want to keep in the main repo? It could also make sense to create a repo just for these inputs/plots, which would pull in a fixed version of `moment_kinetics` as a submodule. If we start making that kind of related-but-separate repo, it might make sense to create a Github organisation for `moment_kinetics` so that we can keep multiple repos grouped together?

PR #155, opened 2023-11-21
Small documentation improvements

johnomotani:
Some better notes on getting started (thanks Lucas!).

Fix some inaccurate descriptions in `test/calculus_tests.jl`, which should not say they use Neumann boundary conditions (which are not even supported in the code) (thanks @mrhardman!).


LucasMontoya4 (approved):
Thank you, this makes it very clear now!

PR #166, opened 2023-12-04
Geometry upgrade

mrhardman:
Generalisation of the treatment of the magnetic geometry to permit the later addition of non-trivial options.

To do:
 - magnetic mirror terms
 - magnetic drift terms
 - vperp advection (d B / dt /= 0)
 - anything else related


mrhardman:
@slnewton @johnomotani https://github.com/mabarnes/moment_kinetics/pull/166/commits/80cca93f388c943b9f740f71dfe1474468a2fd5f my intention is to investigate the toy geometry introduced in the option in the linked commit.

mrhardman:
I have noticed that the distribution function at the initial time is modified as part of the initialisation for manufactured solutions tests. I had previously been careful to make sure that the initial condition was not modified by other routines in the set up by 1) not changing the pdf and 2) making sure that the pdf satisfies the boundary condition. @johnomotani Can you think of any changes in master that might have accidentally modified the initial condition in manufactured solution test cases? This error does not seem to be a show-stopper, but it does make it harder to see where bugs might arise (seeing the fields at the first time step integrated from the exact pdf gives a nice independent test of the integration routines for the manufactured solutions).

johnomotani:
> Can you think of any changes in master that might have accidentally modified the initial condition in manufactured solution test cases?

Not 100% sure, but I think the only thing that the initialisation should do before writing out the initial condition is apply boundary conditions. If it was a moment-kinetic run, the constraints would also be applied, but that shouldn't be affecting your test.

mrhardman:
I am wondering then if I have broken something -- will investigate more carefully!

mrhardman:
The solution is that the d F / dvperp = 0 boundary condition on vperp is responsible for disturbing the manufactured F and causing the observed error in the initial condition.

mrhardman:
[MMS_test_vpa2vperpz_nelement_mirror_wall-1D-2V_fields_and_ion_pdf_no_Er.pdf](https://github.com/mabarnes/moment_kinetics/files/14040235/MMS_test_vpa2vperpz_nelement_mirror_wall-1D-2V_fields_and_ion_pdf_no_Er.pdf)
Trying to make the MMS test with mirror terms, I find the attached errors. Note that the error in the ion distribution function hardly changes with resolution, suggesting a bug. Forgive the shocking placement of the legend (Julia Plots is not very smart, and I have run out of time). @johnomotani I will need to find the cause of the error, but I am running out of ideas. The input files for the test are in the latest commit, in case there is time to try running these cases yourself.

johnomotani:
Are you running the tests in parallel? If so, you might want to try and fix the error that's being picked up by the debug checks - anything can happen if there's an illegal write to memory!

mrhardman:
> Are you running the tests in parallel? If so, you might want to try and fix the error that's being picked up by the debug checks - anything can happen if there's an illegal write to memory!

I am running them in serial, as well as in parallel, with the same results.

To me it looks like there is some kind of instability growing regardless of how one treats the vperp boundary conditions. This might mean that the manufactured solution is not good, or it might mean that the CFL is not satisfied. I have tried changing dt and ngrid for the vperp domain and it seems to make no difference to the final error in the pdf.

mrhardman:
> Are you running the tests in parallel? If so, you might want to try and fix the error that's being picked up by the debug checks - anything can happen if there's an illegal write to memory!

The tests are passing in parallel on my local machine, so I am not sure what is causing the tests to fail on check in.

johnomotani:
> Trying to make the MMS test with mirror terms, I find the attached errors. Note that the error in the ion distribution function hardly changes with resolution, suggesting a bug. Forgive the shocking placement of the legend (Julia Plots is not very smart, and I have run out of time). @johnomotani I will need to find the cause of the error, but I am running out of ideas. The input files for the test are in the latest commit, in case there is time to try running these cases yourself. I do not think that the d F/ d vperp = 0 boundary condition should cause this problem, and if we run without it, some of the higher resolution simulations appear to crash with sqrt(-1).

I started trying to write a test for the dF/dvperp=0 boundary condition. It seems to converge OK, so the difficulty with the MMS tests is weird.

I also tried the MMS tests without applying the dF/dvperp=0 boundary condition, but decreasing the timestep for the nelement=32 case (1D-mirror_MMS_ngrid_9_nel_r_1_z_32_vpa_32_vperp_16_diss.toml) by a factor of 2 - also the 64 element one with dt reduced by 4x, but that was slow enough that I didn't bother to run it long enough to compare - and that test seemed to be converging properly. Could it be the sqrt(-1) error just be CFL condition (although don't ask me why the runs with the dF/dverp=0 boundary condition ran!)?

I noticed the 'vperp advection' speed is proportional to vperp, so it's exactly 0 at vperp=0. Are you sure we need to apply a boundary condition for advection at vperp=0? Maybe we don't as there is exactly zero incoming flux in any case?
If we don't need the boundary condition for advection, do we need to apply the boundary condition on the distribution function array values at vperp=0? In the weak formulation of the collision operator, or of diffusion in vperp, isn't the zero-gradient boundary condition applied by setting the integration-by-parts boundary term as vperp=0 to zero, so maybe that's enough?

PS, I'll push a fix for the debug checks, and one for MMS plots in `makie_post_processing`.

mrhardman:
> I started trying to write a test for the dF/dvperp=0 boundary condition. It seems to converge OK, so the difficulty with the MMS tests is weird.
> 
> I also tried the MMS tests without applying the dF/dvperp=0 boundary condition, but decreasing the timestep for the nelement=32 case (1D-mirror_MMS_ngrid_9_nel_r_1_z_32_vpa_32_vperp_16_diss.toml) by a factor of 2 - also the 64 element one with dt reduced by 4x, but that was slow enough that I didn't bother to run it long enough to compare - and that test seemed to be converging properly. Could it be the sqrt(-1) error just be CFL condition (although don't ask me why the runs with the dF/dverp=0 boundary condition ran!)?

The problem could be that the error is dominated by the error in the timestepper in this example, because the advection velocity is so large (dvperp/dt ~ vpa*vperp ~ 36).

> I noticed the 'vperp advection' speed is proportional to vperp, so it's exactly 0 at vperp=0. Are you sure we need to apply a boundary condition for advection at vperp=0? Maybe we don't as there is exactly zero incoming flux in any case? If we don't need the boundary condition for advection, do we need to apply the boundary condition on the distribution function array values at vperp=0? In the weak formulation of the collision operator, or of diffusion in vperp, isn't the zero-gradient boundary condition applied by setting the integration-by-parts boundary term as vperp=0 to zero, so maybe that's enough?

I have thought about this. Normally in the finite difference codes where we can avoid a boundary condition at vperp = 0, we do use the information that the flux is zero to write the difference operator at vperp = 0, and we avoid including data at vperp = 0. Note that here we do not have a point at vperp = 0 (Radau grid), so the advection term is nonzero for all the vperp points. We are using a pseudospectral method for the advection term, not a "strictly" weak method that has integrated by parts to impose boundary conditions. This choices takes away the ability to specially construct the numerical operator to use that the flux is zero at the origin in vperp. Even in the case that we do use a weak method, in the collision operator or numerical diffusion, I found that a regularity conditon d F / d vperp = 0 was required for long term stability. If you are motivated to show that d F d vperp = 0 is not really required afterall, I would welcome other tests, but I would strongly suspect that the simulations would suddenly blow up.

> PS, I'll push a fix for the debug checks, and one for MMS plots in `makie_post_processing`.

Thanks!

mrhardman:
P.S. to my last comment.

Last week I also tested the `D0` array using the script https://github.com/mabarnes/moment_kinetics/blob/geometry-upgrade/test_scripts/chebyshev_radau_test.jl. This array computes the derivative at coord = 0 (including on grids where coord = 0 is not stored, like vperp). The test seemed to be working well with expected convergence behaviour.

I also ran the MMS test with the d vperp / dt advection manually switched off, but still imposing the vperp boundary condition, and by doing so, I determined that the line where d F / d vperp = 0 is imposed https://github.com/mabarnes/moment_kinetics/blob/13330b30cf2d54367dbe973524acf5c8c3955bbf/src/initial_conditions.jl#L2193 was responsible for the loss of convergence. If this boundary condition was turned off by adding `&& false` to the logical check above L2193, then the simulation appeared to behave much much better in terms of errors observed, even though the time step size was not reduced as coordinate resolution increased. I haven't made a plot to show this, but if it is useful, I can.

mrhardman:
The problems with the MMS test have now been resolved. The issue was a bug in the `D0` matrix, only affecting the Chebyshev discretisation. The figure below shows that the behaviour with mirror physics is now as expected.
[MMS_test_vpa2vperpz_nelement_mirror_wall-1D-2V_fields_and_ion_pdf_no_Er.pdf](https://github.com/mabarnes/moment_kinetics/files/14111222/MMS_test_vpa2vperpz_nelement_mirror_wall-1D-2V_fields_and_ion_pdf_no_Er.pdf)

moment_kinetics/src/time_advance.jl
```diff
@@ -288,11 +288,12 @@ function setup_time_advance!(pdf, vz, vr, vzeta, vpa, vperp, z, r, vz_spectral,
     begin_serial_region()
     vperp_advect = setup_advection(n_ion_species, vperp, vpa, z, r)
     # initialise the vperp advection speed
+    # note that z_advect and r_advect are arguments of update_speed_vperp!
```
johnomotani, line 291: Not sure this comment adds anything. Suggest either remove it, or add an explanation of why it is important to note this.

moment_kinetics/src/time_advance.jl
```diff
@@ -475,13 +478,15 @@ function setup_advance_flags(moments, composition, t_input, collisions,
...
     vz_diffusion = false
     explicit_weakform_fp_collisions = false
     # all advance flags remain false if using operator-splitting
     # otherwise, check to see if the flags need to be set to true
     if !t_input.split_operators
         # default for non-split operators is to include both vpa and z advection together
         advance_vpa_advection = true && vpa.n > 1 && z.n > 1
+        advance_vperp_advection = true && vperp.n > 1 && z.n > 1
         advance_z_advection = true && z.n > 1
         advance_r_advection = true && r.n > 1
```
johnomotani, line 491: Not introduced in this PR, bu the `true &&` looks strange and does not do anything. I think that is just confusing, so we might as well remove.
```suggestion
        advance_vpa_advection = vpa.n > 1 && z.n > 1
        advance_vperp_advection = vperp.n > 1 && z.n > 1
        advance_z_advection = z.n > 1
        advance_r_advection = r.n > 1
```

moment_kinetics/src/z_advection.jl
```diff
@@ -93,10 +93,25 @@ function update_speed_z!(advect, upar, vth, evolve_upar, evolve_ppar, fields, vp
...
+        cvdriftz = geometry.cvdriftz
+        gbdriftz = geometry.gbdriftz
         @inbounds begin
             @loop_r_vperp_vpa ir ivperp ivpa begin
-                @. @views advect.speed[:,ivpa,ivperp,ir] = vpa.grid[ivpa]*bzed + ExBfac*fields.Er[:,ir]
+                # vpa bzed
+                @. @views advect.speed[:,ivpa,ivperp,ir] = vpa.grid[ivpa]*bzed[:,ir]
+                # ExB drift
+                @. geofac = bzeta[:,ir]*jacobian[:,ir]/Bmag[:,ir]
+                @. @views advect.speed[:,ivpa,ivperp,ir] += ExBfac*geofac*fields.Er[:,ir]
```
johnomotani, line 110: This would allocate an unnecessary intermediate array, so do in one line
```suggestion
                @. @views advect.speed[:,ivpa,ivperp,ir] += ExBfac*bzeta[:,ir]*jacobian[:,ir]/Bmag[:,ir]*fields.Er[:,ir]
```

test_scripts/chebyshev_radau_test.jl
```diff
@@ -49,7 +46,8 @@ function chebyshevradau_test(; ngrid=5, L_in=3.0)
     y_nelement_global = y_nelement_local # total number of elements 
     y_L = L_in
     bc = "zero" 
-    discretization = "gausslegendre_pseudospectral"
+    #discretization = "gausslegendre_pseudospectral"
+    #discretization = "chebyshev_pseudospectral"
```
johnomotani, line 50:
```suggestion
```

johnomotani (commented):
Looks fine. One comment to address, then I think we can merge.

johnomotani:
Oh, one other thing - it would be nice to have a docs page (.md file in `docs/src/`) at least briefly describing the available geometry options.

moment_kinetics/src/time_advance.jl
mrhardman: This means that we should make sure to calculate the z_advect and r_advect speeds before we try to use vperp_advect. Is this extra comment enough?

mrhardman:
> Oh, one other thing - it would be nice to have a docs page (.md file in `docs/src/`) at least briefly describing the available geometry options.

Should this file be a new file or should the content go in the zz_geo.md file? Are you happy with the description of the geometry given in the corresponding ExCALIBUR report?

moment_kinetics/src/time_advance.jl
johnomotani: Yes, perfect.

johnomotani:
> Should this file be a new file or should the content go in the zz_geo.md file?

A new file. The files without the `zz_` prefix are hand-written documentation (e.g. https://mabarnes.github.io/moment_kinetics/dev/post_processing_notes/#Post-processing). To add a new one, create the file, and then put the filename in the `Pages` list in `index.md`.

The files with `zz_` prefix are the auto-generated module documentation (from the docstrings in the Julia code). The `zz_` prefix is so that they get listed (in the left-hand sidebar of the docs site) together in a group that comes after all the hand-written pages. For whatever reason that sidebar sorts things by filename, but that turns out to work pretty well for what we want.

> Are you happy with the description of the geometry given in the corresponding ExCALIBUR report?

That'd be great.

mrhardman:
> > Are you happy with the description of the geometry given in the corresponding ExCALIBUR report?
> 
> That'd be great.

Is there a preference for putting the documentation in the geo.jl module next to the source? This seems like a more sustainable option, but perhaps a link in the source to the docs page is enough to ensure it is read and modified when the source is updated?

johnomotani:
> Is there a preference for putting the documentation in the geo.jl module next to the source? This seems like a more sustainable option, but perhaps a link in the source to the docs page is enough to ensure it is read and modified when the source is updated?

That is an option, and I agree that it's nice to keep the source code and docs together. The downside is that it's a little bit more annoying to write documentation in docstrings in .jl files than in a .md file (e.g. when writing maths, the `$` characters all need to be escaped, and I think some other similar things). So whichever you like. If you decide to put all the docs into the source code file, it would still be nice to have a `geo.md` page that can be linked in the index that just links to the module docs (the link would be created by ``[`moment_kinetics.geo`](@ref)``).

mrhardman:
OK. I am happy with linking to external libraries. For now the main problem is that I cannot work out how to make or view the documentation locally without working on git. Is it really necessary to have webviewer? That may be a problem, as I do all my development on remote servers.

johnomotani:
PS the README.md for the docs might need to be updated. I think you need some setup before the docs will build along the lines of
```
$ cd docs/
$ julia --project
julia>]
pkg> dev ../moment_kinetics
pkg> dev ../makie_post_processing/makie_post_processing
pkg> dev ../plots_post_processing/plots_post_processing
```
After that (I hope!)
```
julia> include("make.jl")
```
or
```
$ julia --project make.jl
```
should work. As usual, I prefer the REPL version because when you need to edit the .md files and rebuild the docs, the second (and later) docs builds are fast because the code is already compiled.

mrhardman:
I have just added a new file https://github.com/mabarnes/moment_kinetics/blob/geometry-upgrade/docs/src/geometry.md. Please comment!

Whilst writing this I considered that an input namelist for the geometry option (like numerical dissipation) would be appropriate. @johnomotani do you agree and should this happen in this PR?

johnomotani:
> Whilst writing this I considered that an input namelist for the geometry option (like numerical dissipation) would be appropriate. @johnomotani do you agree and should this happen in this PR?

I think that should happen in general. This seems like a good time to do it for the geometry. Including the change in this PR would be good as we are already breaking the input options (removing `Bzed`, etc.) so better to finish breaking them before people start writing new input files.

Once you create the new section, it would be good to add the 'old' top-level options to `removed_options_list` (see https://github.com/mabarnes/moment_kinetics/pull/166/commits/b52f9430a009702b61c1eb065bbb3d793bf7fc27) so people don't get surprised if old input files don't behave as expected.

johnomotani:
> I have just added a new file https://github.com/mabarnes/moment_kinetics/blob/geometry upgrade/docs/src/geometry.md. Please comment!

Nice, thank you!

mrhardman:
> > Whilst writing this I considered that an input namelist for the geometry option (like numerical dissipation) would be appropriate. @johnomotani do you agree and should this happen in this PR?
> 
> I think that should happen in general. This seems like a good time to do it for the geometry. Including the change in this PR would be good as we are already breaking the input options (removing `Bzed`, etc.) so better to finish breaking them before people start writing new input files.
> 
> Once you create the new section, it would be good to add the 'old' top-level options to `removed_options_list` (see [b52f943](https://github.com/mabarnes/moment_kinetics/commit/b52f9430a009702b61c1eb065bbb3d793bf7fc27)) so people don't get surprised if old input files don't behave as expected.

This is now done, please review my latest changes!

johnomotani (approved):
Looks good to me. We can merge when the CI tests pass.

PR #170, opened 2023-12-13
Precompile on rank-0 first in run_moment_kinetics.jl

johnomotani:
When running from the command line using the `run_moment_kinetics.jl` script, call `using moment_kinetics` first on rank-0, and only after that has completed call it on all other ranks. Hopefully this avoids precompilation clashes (when multiple MPI ranks try to precompile the same package simultaneously) which might corrupt files. It does not solve the clash issue for importing MPI, because `using MPI` is needed to get the MPI commands to synchronize between rank-0 and the other processes, but hopefully minimises the chance of problems.

### DIFF
--- a/run_moment_kinetics.jl
+++ b/run_moment_kinetics.jl
@@ -2,6 +2,19 @@
 using Pkg
 Pkg.activate(".")
 
-using moment_kinetics
+using MPI
+MPI.Init()
+is_rank0 = (MPI.Comm_rank(MPI.COMM_WORLD) == 0)
+
+# Use MPI Barriers to ensure `using moment_kinetics` is completed on rank-0 (in order to
+# complete precompilation of moment_kinetics and dependencies) before being run on all the
+# other MPI ranks.
+if is_rank0
+    using moment_kinetics
+    MPI.Barrier(MPI.COMM_WORLD)
+else
+    MPI.Barrier(MPI.COMM_WORLD)
+    using moment_kinetics
+end
 
 run_moment_kinetics()


johnomotani:
An alternative (or additional) possible fix would be to add a line to submission scripts, e.g. if your command to run a simulation is
```
mpirun -np $SLURM_NTASKS julia --project -O3 --check-bounds=no run_moment_kinetics $INPUT_FILE
```
you could add a line, modifying it to
```
julia --project -O3 --check-bounds=no -e "using Pkg; Pkg.precompile()"
mpirun -np $SLURM_NTASKS julia --project -O3 --check-bounds=no run_moment_kinetics $INPUT_FILE
```
(It is important that the flags, here `-O3 --check-bounds=no` are the same between the two commands to avoid repeated precompilation.)

This alternative might actually speed things up a bit as `Pkg.precompile()` should run the precompilation in parallel, whereas I think the implicit precompilation when `using moment_kinetics`, etc., is called only happens in serial.

mrhardman:
In order to deal with some of my precompile issues caused by working on multiple instances of moment_kinetics simultaneously, I have started using the command
```
cd path/to/moment_kinetics; JULIA_DEPOT_PATH=$(pwd); julia --project
```
to make sure that the .julia files are installed in an encapsulated location for each separate development project folder.

mrhardman:
There seem to be many related discussions of similar issues on the Julia discourse. Noting here for future reference.

https://discourse.julialang.org/t/how-does-one-set-up-a-centralized-julia-installation/13922
https://discourse.julialang.org/t/run-a-julia-application-at-large-scale-on-thousands-of-nodes/23873/3
https://discourse.julialang.org/t/precompilation-error-using-hpc/17094
https://discourse.julialang.org/t/repeated-precompilation-on-a-cluster-makes-life-difficult/94194/12

etc...

PR #174, opened 2024-01-02
Make some dependencies optional, reorganise package structure, improve setup scripts

johnomotani:
Some dependencies are heavy, and take a long time to precompile, in particular `Symbolics`, `Makie` and `Plots`. `Symbolics` is only needed to run MMS tests, and only one of `Makie` or `Plots` is needed. This PR allows all these dependencies to be optional.

To allow making dependencies optional, the structure of the package (now packages) is reorganised. `moment_kinetics`, `makie_post_processing`, and `plots_post_processing` are now separate packages, contained in subdirectories. To use `moment_kinetics` a project is created in the top-level directory of the repo into which some or all of the packages and optional dependencies can be installed.

To simplify the setup of `moment_kinetics` (especially since all the optional things now make it more complicated), the setup script in `machines/machine_setup.sh`, etc., is updated and upgraded. It now supports setup on a laptop/desktop (the 'generic-pc' target), has a template setup for a 'generic-batch' machine that can be used to support HPC machines other than the explicitly included ones, and has some extra features too.

For more info, see the updated documentation:
https://mabarnes.github.io/moment_kinetics/previews/PR174/getting_started/#Getting-started
https://mabarnes.github.io/moment_kinetics/previews/PR174/developing/#Package-structure
https://mabarnes.github.io/moment_kinetics/previews/PR174/manual_setup/#Manual-setup
https://mabarnes.github.io/moment_kinetics/previews/PR174/machine_setup_notes/


makie_post_processing/makie_post_processing/src/shared_utils.jl
mrhardman, line 1: The fact that shared utils appear in the makie_* directory makes it seem that the plots_post_processing will need the make_post_processing to compile. Is that correct? If so, can we put this file somewhere else?

mrhardman:
The current package structure seems to require that python packages are installed even if one uses the "Plots" based post processing module.
``` 
pkg> develop ./plots_post_processing/plots_post_processing/
    ... lots of output
    Building Conda ─→ `~/excalibur/moment_kinetics_geometry/.julia/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/51cab8e982c5b598eea9c8ceaced4b58d9dd37c9/build.log`
    Building PyCall → `~/excalibur/moment_kinetics_geometry/.julia/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/9816a3826b0ebf49ab4926e2b18842ad8b5c8f04/build.log`
```

makie_post_processing/makie_post_processing/src/shared_utils.jl
johnomotani: No, I just used an ugly relative path from `plots_post_processing.jl`:
https://github.com/mabarnes/moment_kinetics/blob/1f0f3d01259166eb52ae3102fd06f58c250596ce/plots_post_processing/plots_post_processing/src/plots_post_processing.jl#L18

johnomotani:
> The current package structure seems to require that python packages are installed even if one uses the "Plots" based post processing module.

Some of the plots use PyPlot (specifically the plots of the moment-kinetic distribution function using the unnormalised velocity coordinates as axes, as Plots.jl does not support 2d x- and y-positions). It would be technically possible to make that an optional dependency if someone wants to... but if it's causing you a problem I'd suggest hacking - just comment out the PyPlot/PyCall bits in `plots_post_processing/plots_post_processing/src/plots_post_processing.jl` and `plots_post_processing/plots_post_processing/Project.toml`.

johnomotani:
@mrhardman @mabarnes I think this is ready to merge now. Do you think there's anything left to fix or document better?

PR #178, opened 2024-02-01
Merge separate-postprocessing-packages into #166

johnomotani:


mrhardman:
I started attempting this myself here https://github.com/mabarnes/moment_kinetics/tree/geometry-upgrade-separate-packages.

The automatic CI tests on that branch did compile and run on Julia 1.9.3

The first thing I can see in conflict with the way I had to introduce geometry was that I had to split up the "get_geometry_and_composition" function because I didn't want to make the MMS test scripts unnecessarily reproduce the geometry arrays. There is a lot of nice things to be said for using the same functions in post processing as in the main code, but it is also a pain if one has to inherit a lot of data that isn't available to the script.

johnomotani:
> The first thing I can see in conflict with the way I had to introduce geometry was that I had to split up the "get_geometry_and_composition" function because I didn't want to make the MMS test scripts unnecessarily reproduce the geometry arrays. There is a lot of nice things to be said for using the same functions in post processing as in the main code, but it is also a pain if one has to inherit a lot of data that isn't available to the script.

Ah, sorry, I'll add that back - I missed that `get_composition()` is actually used on its own in some places.

Anyway, there weren't too many conflicts, so hopefully either version of the merge should be fairly easy to polish up.

johnomotani:
#179 looks good to me. I thin we should just merge that into #166, so closing this.

PR #184, opened 2024-02-23
Make machine setup scripts work on MacOS

### LINKED ISSUE #183: machine setup fails on macbook
I am running on a Mac, and In machine setup options, I obtain the following error when entering the julia executable path.

Please enter the path to the Julia executable to be used ['/opt/local/bin/julia']:
> /opt/local/bin/julia
realpath: illegal option -- s
usage: realpath [-q] [path ...]

I can install manually but would be nice to get the script working.  Maybe I'm not understanding something I should be doing?

johnomotani:
Replace `realpath` with Python in machine setup script. `realpath` is not necessarily portable, while we already require Python to be available for the Julia-downloader script. Using Python's `os` module should be pretty portable.

Fixes #183.

### DIFF
--- a/machines/generic-batch-template/compile_dependencies.sh
+++ b/machines/generic-batch-template/compile_dependencies.sh
@@ -79,7 +79,7 @@ fi
 if [[ $BUILDHDF5 == "y" ]]; then
   HDF5=hdf5-1.14.3
   # Download and extract the source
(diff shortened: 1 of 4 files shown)


johnomotani:
@mabarnes could you see if this PR fixes the issue in #183 please?

mabarnes:
i've just pulled the latest source and still get the same error referencing realpath... :(

johnomotani:
@mabarnes are you using the changes from this PR (which hasn't been merged into anything else yet)? The `realpath` command line shouldn't be being called at all!

Issue #186, opened 2024-02-28
Gyroaverages


mrhardman:
Gyroaverages may be crucial for providing a physical cut-off to the wavenumber spectrum required to be resolved in the moment_kinetics model.

Initial experiments implementing an ion gyroaverage is in the branch https://github.com/mabarnes/moment_kinetics/tree/feature-gyroaverages.

To test the feature, run https://github.com/mabarnes/moment_kinetics/blob/feature-gyroaverages/test_scripts/gyroaverage_test.jl.

Here the Lagrange Polynomial representation is used to precompute integration weights as a function of z and r, in place of the integral in gyrophase, which is only a line integral in (z,r) space. The test plots the field `phi` and the gyroaverage `gphi` for several values of `vperp`. Boundary conditions are not yet dealt with, meaning that the code will definitely produce poor results near the domain boundary. The code is written assuming fully local grids in memory.

To do list:
- [x] Implement sandbox-gyroaverage
- [x] Create a test for the gyroaverage feature
- [ ] Determine good ideal behaviour for boundary regions
- [ ] Implement good boundary behaviour
- [x] Implement the gyroaverage with shared-memory MPI 
- [ ] Consider how to efficiently implement the gyroaverage with distributed memory MPI
- [x] Consider how to leverage the sparse nature of the gyroaverage operator to speedup evaluation
- [x] Extend gyroaverage feature to main code
- [ ] Compare to numerical dissipation
Comments appreciated @johnomotani @mabarnes.

mrhardman:
This question seems to be partially address here by this thread discussing the construction of a vector of vectors of variable length: https://discourse.julialang.org/t/q-vector-of-vectors-of-variable-lengths/102817. By first looping over the `gyromatrix` to find and count the nonzero elements, we can construct an array `sizes` that has shape `(nvperp,nz,nr,nspecies)` and the value of which is the number of nonzero `(izp,irp)` points for that part of phase space. We can convert `sizes` into an array of vectors via the constructor method
```
array_of_vecs = zeros.(Int64,sizes)
```
and then fill with the required index `iz` or `ir`. Two of these arrays will be required for simplicity.

mrhardman:
[...]

PR #187, opened 2024-02-29
Feature gyroaverages

mrhardman:
To do list:
- [x] Implement sandbox-gyroaverage
- [x] Create a test for the gyroaverage feature
- [ ] Determine good ideal behaviour for boundary regions
- [ ] Implement good boundary behaviour
- [x] Implement the gyroaverage with shared-memory MPI 
- [ ] Consider how to efficiently implement the gyroaverage with distributed memory MPI
- [x] Consider how to leverage the sparse nature of the gyroaverage operator to speedup evaluation
- [x] Extend gyroaverage feature to main code
- [ ] Compare to numerical dissipation


mrhardman:
[periodic_box_low_res.zip](https://github.com/mabarnes/moment_kinetics/files/14602112/periodic_box_low_res.zip)
Low resolution runs with `rhostar = 0.1` and box lengths of 1 in r and z. Neither run uses radial diffusion, but the run `2D-periodic-gk.toml` uses a gyroaverage. Clear differences can be seen where the gyrokinetic run tends to have smoother features and smaller amplitudes in the peaks and troughs of the solution, whereas the run `2D-periodic-dk.toml` obtains grid scale behaviour. A careful study of this 2D model might be useful for understanding the slab ITG instability. @johnomotani @mabarnes @LucasMontoya4. Further efficiency improvements in the GK operator are likely required to go to higher resolutions.

mrhardman:
Marking this PR as ready for review as shared-memory MPI is now working, looking for feedback before implementing distributed-memory MPI support. @johnomotani @mabarnes

mrhardman:
Commit https://github.com/mabarnes/moment_kinetics/pull/187/commits/cd85aa7ae8cdc5bf66340eb726dabe6238e74a1c passes all tests apart from tests in parallel on MacOS, which mysteriously time out after seeming to finish with the tests reportedly passing (although a different number of tests seem to be carried out as compared to the Ubuntu case.

johnomotani:
> Commit cd85aa7 passes all tests apart from tests in parallel on MacOS, which mysteriously time out after seeming to finish with the tests reportedly passing (although a different number of tests seem to be carried out as compared to the Ubuntu case.

The macOS parallel tests do this fairly frequently. Often just re-running the jobs lets them pass, so I've not investigated further.

The macOS parallel tests were being slow, so I cut them down to 1 test run (4 process, no `--long`). On the Linux job, we test 4 processes, 3 processes, and finally 2 processes. On the 2-processes test we use the `--long` flag - that's why there will be some more tests reported for that run than there are in the macOS job.

moment_kinetics/src/time_advance.jl
```diff
@@ -1441,23 +1425,65 @@ end
...
+    if composition.gyrokinetic_ions
+        # fill buffer with ring-averaged F (gyroaverage at fixed position)
+        gyroaverage_pdf!(ff,new_scratch.pdf,gyroavs,vpa,vperp,z,r,composition)
+    else
+        # copy F into buffer (drift-kinetic)
+        begin_s_r_z_vperp_vpa_region()
+        @loop_s_r_z_vperp_vpa is ir iz ivperp ivpa begin 
+            ff[ivpa,ivperp,iz,ir,is] = new_scratch.pdf[ivpa,ivperp,iz,ir,is]
+        end
+    end
```
johnomotani, line 1441: Can skip the copy when not using gyroaverage
```suggestion
    if composition.gyrokinetic_ions
        ff = scratch_dummy.buffer_vpavperpzrs_1
        # fill buffer with ring-averaged F (gyroaverage at fixed position)
        gyroaverage_pdf!(ff,new_scratch.pdf,gyroavs,vpa,vperp,z,r,composition)
    else
        ff = new_scratch.pdf
    end
```

moment_kinetics/src/time_advance.jl
```diff
@@ -867,21 +877,16 @@ function time_advance!(pdf, scratch, t, t_input, vz, vr, vzeta, vpa, vperp, gyro
...
         t += t_input.dt
 
-        if i == t_input.nstep
-            # Ensure all output is written at the final step
-            finish_now = true
-        end
-
-        if mod(i,t_input.nwrite_moments) == 0 || mod(i,t_input.nwrite_dfns) == 0 || finish_now
+        if diagnostic_checks
             # update the diagnostic chodura condition
```
johnomotani, line 888:
```suggestion
```

moment_kinetics/src/time_advance.jl
```diff
@@ -867,21 +877,16 @@ function time_advance!(pdf, scratch, t, t_input, vz, vr, vzeta, vpa, vperp, gyro
...
-            update_chodura!(moments,scratch[end].pdf,vpa,vperp,z,r,spectral_objects.r_spectral,composition,geometry,scratch_dummy,advect_objects.z_advect)
+            # update_chodura!(moments,scratch[end].pdf,vpa,vperp,z,r,spectral_objects.r_spectral,composition,geometry,scratch_dummy,advect_objects.z_advect)
```
johnomotani, line 890: Calculation has moved to `update_derived_moments!()`, so delete commented-out version.
```suggestion
```

moment_kinetics/src/velocity_moments.jl
```diff
@@ -486,8 +486,22 @@ end
...
+    if composition.gyrokinetic_ions
+        # fill buffer with ring-averaged F (gyroaverage at fixed position)
+        gyroaverage_pdf!(ff,ff_in,gyroavs,vpa,vperp,z,r,composition)
+    else
+        # copy F into buffer (drift-kinetic)
+        begin_s_r_z_vperp_vpa_region()
+        @loop_s_r_z_vperp_vpa is ir iz ivperp ivpa begin 
+            ff[ivpa,ivperp,iz,ir,is] = ff_in[ivpa,ivperp,iz,ir,is]
+        end
+    end
```
johnomotani, line 504: Skip copy when not gyroaveraging
```suggestion
    if composition.gyrokinetic_ions
        ff = scratch_dummy.buffer_vpavperpzrs_1 # the buffer array for the charged pdf -> make sure not to reuse this array below
        # fill buffer with ring-averaged F (gyroaverage at fixed position)
        gyroaverage_pdf!(ff,ff_in,gyroavs,vpa,vperp,z,r,composition)
    else
        ff = ff_in
    end
```

johnomotani (approved):
I haven't checked the actual gyroaverage functionality or tests, but lets merge this to get the updates to moment-update functions into `master`.

Gyroaverage feature is still a prototype, we should open an issue to track the remaining todos.

mrhardman:
This issue can act as the tracker: https://github.com/mabarnes/moment_kinetics/issues/186

mrhardman:
It looks like some of the parallel tests might have been broken, rerunning the tests as I don't understand how that can be.

johnomotani:
It looks like parallel tests sometimes (often?) fail and sometimes pass, which is a bit suspicious. It would be good to add a 'debug check' that uses the gyroaverage - I'll have a look at that now.

johnomotani:
Ah, the PR tests (in parallel) were failing because of changes from #203 (which had been merged into `master`. I've fixed that now (I think), but the CI runners are failing to install OpenMPI. I don't think that failure is anything to do with us, so hopefully it will sort itself out in a little while.

PR #192, opened 2024-03-13
First pull request LM_sandbox to electrons-jto-debug

LucasMontoya4:


examples/kinetic-electrons/wall+sheath-bc_boltzmann_loworder.toml
```diff
@@ -49,10 +49,10 @@ ionization_frequency = 2.0
 #electron_ionization_frequency = 2.0
 #ionization_energy = 1.0
 constant_ionization_rate = false
-nstep = 1000000
+nstep = 1000
 #nstep = 1
 dt = 1.0e-5
-nwrite = 10000
+nwrite = 100
```
johnomotani, line 55: Lets revert this change too
```suggestion
nwrite = 10000
```
and (if you see this comment in time) a nice thing to do with similar/related suggestions is to group them together with the 'Add suggestion to batch' button, because that makes a single commit rather than several (which makes the history a little bit tidier).

examples/kinetic-electrons/wall+sheath-bc_boltzmann_loworder.toml
```diff
@@ -85,8 +85,8 @@ ascii_output = true
 [numerical_dissipation]
 #moment_dissipation_coefficient = 0.0001
 #moment_dissipation_coefficient = 1.0
-vpa_dissipation_coefficient = 0.002
-#vpa_dissipation_coefficient = 0.1
+#vpa_dissipation_coefficient = 0.002
+ion_vpa_dissipation_coefficient = 0.1
```
johnomotani, line 89: You don't have to do anything for this comment, but to explain why in case you wonder about updating all the places `vpa_dissipation_coefficient` is used in example input files, etc.: that is the one other thing we would need to do for this change (i.e. update all the uses of `vpa_dissipation_coefficient` in the example input files and the tests). It's not useful to actually do that right now, since you're working on changing the structure like we discussed yesterday - once you've done that the change to the inputs will be to replace `[numerical_dissipation]` with `[ion_numerical_dissipation]` rather than `vpa_dissipation_coefficient` with `ion_vpa_dissipation_coefficient`, so it'd be a waste of time to change all the places where `vpa_dissipation_coefficient` appears in inputs, just to change them back again.

johnomotani (commented):
Looks good, thanks @LucasMontoya4! Just a couple of tiny comments to deal with :+1:

examples/kinetic-electrons/wall+sheath-bc_boltzmann_loworder.toml
LucasMontoya4: I didn't see the notifications for these comments when you first sent them, sorry!! I hadn't realised that comments can happen here too...

Luckily, I have made sure not to do what you said to avoid, because of the future changes. I've committed your suggestions (separately, but now I see how to do the batch suggestions), and now I'm trying to run tests on my local updated code.

johnomotani (approved):
Good job @LucasMontoya4! All neat and tidy :+1: Thanks for updating all the example input files, etc. too!

Very tiny suggestions below to make a couple of input files more consistent and fix a couple of docstrings. I'll commit those changes in a minute, and merge this.

LucasMontoya4:
Thanks John! Sorry about the docstrings.. Also, I don’t know if this matters much but I changed the name of the `calculate_moment_derivatives` functions for each species, because the naming system was different for each. I made sure to alter the names elsewhere in the other files and I guess it must be all fine because the tests all pass?

johnomotani:
>  I changed the name of the calculate_moment_derivatives functions for each species

I saw that, it seemed like a good idea to me - makes the names more consistent and easier to read.

PR #201, opened 2024-04-11
In machine_setup.sh on macOS, prompt for MPI library path

johnomotani:
`MPIPreferences.use_system_binary()` does not find the MPI library when it was installed with homebrew on macOS, so prompt for the user to give the path to the MPI library instead.

@LucasMontoya4 does this work for you?

@mabarnes did you need to work around this problem when using the `machines/machine_setup.sh` setup script?

Is there a default that we should check (for non-homebrew setups?) that could find the MPI library automatically? I don't have a Mac to test...

### DIFF
--- a/machines/shared/add_dependencies_to_project.jl
+++ b/machines/shared/add_dependencies_to_project.jl
@@ -96,15 +96,21 @@ elseif Sys.isapple()
     # path instead.
     # ?? Could we attempt to auto-detect the MPI library before prompting the user??
     if prompt_for_lib_paths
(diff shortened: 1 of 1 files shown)


LucasMontoya4:
@johnomotani I’ll try that one out - although I’m not sure how common it is for Mac users to install MPI with homebrew, but I guess it’s still a general solution!

johnomotani:
I've added the try-catch so if `MPIPreferences.use_system_library()` (with no extra arguments) runs without error, then we don't need to prompt for the MPI library path. So I think this is ready to merge now.

PR #202, opened 2024-04-11
Refactor collision input options to make separate namelist for Krook and Fokker Planck collision operators.

mrhardman:
Changes to permit the collision operator inputs to be kept in separate structs.

Making these changes allowed me to review the formulation for the inputs. To simplify the logic for specifying default behaviour, I introduce new "use_krook" and "use_fokker_planck" parameters that must be true for the operators to be used. Checks on use_krook and use_fokker_planck are done when setting up the advance Bools. The "frequency_option" is used to choose whether or not the collision frequency is specified manually or from reference parameters. I think that this formulation is more natural, but this may be a matter of taste. My concern with the previous formulation was that the same string use used to determine the way the frequency is calculated as well as whether or not the operator is used at all.

A by-product of these changes is that a collision frequency from reference parameters for the Fokker-Planck operator is now available.

The feature should now mean that defaults are only specified in a single location, and that further parameters can be easily added in an encapsulated way.

Feedback appreciated.


moment_kinetics/src/fokker_planck.jl
```diff
@@ -68,6 +70,46 @@ using ..fokker_planck_test: Cssp_fully_expanded_form, calculate_collisional_flux
...
+"""
+function setup_fkpl_collisions_input(toml_input::Dict, reference_params)
+    # get reference collision frequency (note factor of 1/2 due to definition choices)
+    nuii_fkpl_default = 0.5*get_reference_collision_frequency(reference_params)
+    # read the input toml and specify a sensible default
+    input_section = set_defaults_and_check_section!(toml_input, "fokker_planck_collisions",
+       # begin default inputs (as kwargs)
+       use_fokker_planck = false,
+       nuii = -1.0,
+       frequency_option = "manual")
```
johnomotani, line 90: I think the Krook collisions and FP collisions should have the same default setting for `frequency_option`, either both `"manual"` or both `"reference_parameters"`. At the moment Krook has `"reference_parameters"` and FP has `"manual"`.

Should we vote on which to choose? I'd vote for `"reference_parameters"` as the default, to make it easier to have self-consistent inputs. I guess @mrhardman would vote the other way!

mrhardman: I would prefer manual input as default, but I can understand your reasoning. Let's change the default to `"reference_parameters"` on the basis that I will remember to ask for the `"manual"` option.

johnomotani (approved):

PR #217, opened 2024-05-21
MMS test bugfixes and docs

mrhardman:
PR to add 

- [ ] sufficient documentation to make the existing manufactured solutions tests more accessible to new users @LucasMontoya4 
- [ ] documentation of the existing manufactured solution input files (see here for now https://github.com/mabarnes/moment_kinetics/blob/mms_bugfixes_and_docs/docs/src/manufactured_solution_test_examples.md, the link the README.md of the main project does not work yet)
- [ ] (optional on request on from @LucasMontoya4) documentation of how to use the existing plotting functions
- [ ] (optional on request on from @LucasMontoya4) minor refactoring of the plotting functions to make them more user friendly

In addition, this could be a moment to discuss the best location to store the existing MMS input files - > should they be moved from runs to an examples_mms/ directory? @johnomotani


mrhardman:
If I understand correctly, one consideration for the location of the MMS test TOML inputs could be that we might want to test whether the inputs are valid on CI, by running them for 1 timestep, like the TOML in examples/. Currently it looks like many of the MMS test inputs in the `runs/` directory have become out of date following the various input namelist changes.

johnomotani:
BTW all the documentation builds seem to be broken at the moment - there is some error about a missing `get_cmap` function when loading PyPlot. I don't think it's anything we've done, and I can't reproduce it on my laptop, so I'm hoping it goes away with some package updates (unless anyone wants to dig into it, e.g. excluding the version of PyPlot that it's currently trying to use so it falls back to the previous one?).

johnomotani (approved):

PR #218, opened 2024-05-27
Add Maxwellian away from walls in initial condition for wall-bc sims

johnomotani:
Helps to avoid double-peaked distribution function, so helps to avoid instability during initial transient.

Edit: Size of the Maxwellian added is controlled by the `z_IC_density_amplitude1` parameter.


johnomotani:
Note: Size of the Maxwellian added is controlled by the `z_IC_density_amplitude1` parameter.

Should also add some input files and READMEs showing how to initialise the examples, but that can wait for another PR.

PR #219, opened 2024-05-28
Initial functionality for partially-implicit 'IMEX' timestepping

johnomotani:
Runge-Kutta timestepping schemes that can handle terms split into an 'implicit' group and and 'explicit' group are known as 'IMEX' schemes. A couple of examples are introduced in this PR, from [Kennedy & Carpenter 2003](https://doi.org/10.1016/S0168-9274(02)00138-1) and [Kennedy & Carpenter 2019](https://doi.org/10.1016/j.apnum.2018.10.007).

The IMEX schemes are built up from linear combinations of forward-Euler (explicit) and backward-Euler (implicit) steps. To implement the backward-Euler steps (added in this PR), we add a 'Jacobian-free Newton-Krylov' iterative solver (e.g. [Knoll & Keyes 2004](https://doi.org/10.1016/j.jcp.2003.08.010)).

Current examples of choices for the 'implicit terms' are: i) just ion 'vpa advection' in the ion kinetic equation; ii) all terms in the ion kinetic equation (excluding the ion moment equations in the moment-kinetic case). These choices seem to be not that useful - the CFL limits for different terms (in particular the smallest two in a 1D1V test case including wall boundary conditions are the 'ion vpa advection' and 'neutral vz advection') are not that far apart, so the implicitness being added does not allow a huge increase in timestep (in the best case, would be about a factor of 2x to 4x). Unfortunately, the IMEX RK scheme cannot exceed the CFL limit by as large a factor as our 'standard' 4-stage 3rd-order explicit scheme, so we do not end up with a longer timestep than the best explicit scheme we have.

However, the motivation for introducing this implicit functionality is to apply it to the electron equation (once that is added...). The code in this PR provides a sufficient framework for that, so I propose to merge now as-is. Useful extensions to look at one day for use with the ion/neutral equations could be looking at different combinations of terms, and experimenting with preconditioners. The option to apply a function implementing a preconditioner is included, but no preconditioners are provided yet in this PR.

Some documentation is needed. I'd like to merge this PR anyway, as there are structural changes to the timestepping. If people are OK with that, I'll open an issue to flag that documentation is needed (when I get time...).


mrhardman:
If you think this PR is mission critical for the electron implementation then I am happy for you to go ahead once the CI tests pass and you have checked one of the MMS tests still run.

Question: If the ion kinetic equation is supported, does this mean that the Fokker-Planck implementation is also now IMEX with this PR? A good way to justify these changes in a report would be to use the IMEX collision operator and demonstrate that much longer timesteps are possible with a fine velocity resolution. I have been concerned about this problem since seeing how small a timestep was required for the fast-ion-electron collisions in the slowing down relaxation problems that we considered recently.

Question: how hard is it to extend the IMEX stepper to the moment equations? One interest in making the moment equations implicit would be to allow (in the distant future) for electromagnetic fields, for which the other finite-element Full-F/Full solution codes (like JOREK) seem to use a fully implicit solver. 

In general documentation and more clarity is needed in the timestepping area of the code, so I am happy for documentation to be delayed until you are happy with the functionality.

PR #228, opened 2024-07-11
A series of useful python scripts based on xarray for plotting data f…

mrhardman:
…rom moment_kinetics simulations.

These scripts are useful for making specialist plots or publication ready plots with matplotlib. Comments or suggestions on the best way to record these scripts would be useful.


mrhardman:
> Only skimmed very quickly, so sorry if these are stupid questions - are these scripts to make specific plots for your paper? There seem to be quite a few hard-coded paths, etc. If so, I'd suggest making a subdirectory in the `publication_inputs/` directory, and sticking the scripts and maybe a README in there.

Yes these scripts are for making specific plots, with a lot of assumptions. I thought that putting them here would be the best way to guarantee reproducibility in the long term. Happy to structure the scripts as you like. The actual data requires some heavy simulations. Would you ask that the README describes how to make the data for the scripts? What else?
 
 
> `xarray` is a useful and very powerful package (although with a steep learning curve if you're used to numpy/matplotlib!), I'm a fan, but in this PR it looks like you only use it in a way that could be replaced directly with `h5py`, which would probably be marginally faster. Is there a reason for using `xarray` here?

The reason is that I was trying to learn xarray, but indeed, I just used it to extract data as I found the plotting functions to difficult to understand with my matplotlib (pyplot) background. The time limiting step here is computing the error norms of the pdf, not the loading of the data. I think I used h5py for the saving of intermediate data in these scripts, so I could change to h5py everywhere, but only if there are strong opinions on this.

PR #231, opened 2024-07-24
Improve and document 'recycling fraction' example inputs

johnomotani:
Make Fekete4(3) default timestep method in recycling fraction examples. Also add non-moment-kinetic examples, use adaptive timestep for the 'init' runs too, fix Krook collisions options, and add a README.md.


johnomotani:
@asnodin, could you see if these inputs work for you (see the README.md added in this PR)? Thanks!

PR #233, opened 2024-08-02
Allow printing of @enum values to TOML as Strings

johnomotani:
Adds a method of `TOML.Internals.Printer.to_toml_value` that handles `Enum` values, allowing `Enum`s to be printed to TOML format strings or files.

Counterpart to the `get(d::Dict, key, default::Enum)` that lets us read `Enum`s from `Dict`s, which means we can read them from TOML files by supplying a value of the `Enum` as the default.

Slightly hacky because `TOML.Internals.Printer.to_toml_value` appears to be an internal method, not part of the public API, but `TOML.jl` does not change very much, so in practice this should be fine.

### DIFF
--- a/moment_kinetics/src/input_structs.jl
+++ b/moment_kinetics/src/input_structs.jl
@@ -633,6 +633,13 @@ function get(d::Dict, key, default::Enum)
     end
 end
 
+using TOML
+import TOML.Internals.Printer: to_toml_value
+# Define a method of TOML.to_toml_value that handles Enum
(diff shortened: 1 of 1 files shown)


johnomotani:
On second thoughts, this is a bad solution. Modifying `TOML.Internals.Printer.to_toml_value()` means that we would change the behaviour of `TOML.print()` (i.e. how it handles `Enum`) as a side effect whenever `moment_kinetics` is loaded. I think a better solution is to have our own function for writing to TOML - see upcoming PR.

PR #236, opened 2024-08-07
Function `options_to_TOML()` adds special handling for `Enum`

johnomotani:
Replaces #233.


johnomotani:
This PR now also handles `Enum` in `set_defaults_and_check_section!()`, which should complete the handling of `Enums` for input (they can now be set also in sub-sections of the input), and in output (not supported before this PR).

PR #244, opened 2024-09-04
Fix get_run_info_no_setup when n_neutrals = 0

LucasMontoya4:


johnomotani:
Thanks @LucasMontoya4!

PR #245, opened 2024-09-06
Solver for  Poisson's equation in (radial,polar) coordinates.

mrhardman:
This PR gives the option of solving 

```math
  \frac{1}{r} \frac{\partial }{\partial r}\left( r \frac{\partial \phi}{\partial r}\right) + \frac{1}{r^2}\frac{\partial^2 \phi}{\partial \theta^2} = \rho(r,\theta)
```
for $\phi = \phi(r,\theta)$ using two different methods.

First, Fourier transforms for $\theta$ and Gauss-Legendre finite-element methods in 1D for the resulting system of independent ODEs. A Fourier spectral object (separate from Chebyshev) is added. Dirichlet (zero) boundary conditions are imposed on $\phi$ at the radial boundary. In the test, the radial domain runs from $(0,L]$.

Second, a 2D tensor-product Gauss-Legendre finite-element method, using the functions originally designed for the Fokker-Planck collision operator. Some base level functions have to be moved to effect this change. The sparse-matrix version of the assembly where the periodic boundary condition is imposed is currently broken, so for this PR we resort to the dense-matrix version of the operator creation (for now).

These features (and their extensions) may be useful if more complicated field solvers are eventually required in moment_kinetics, for example, for solving for $\phi$ from vorticity on closed field lines, or solving for the electromagnetic fields.

An automatic test is added which runs in serial (even if multiple cores are passed).


mrhardman:
Closed as not aligned with other development.

PR #246, opened 2024-09-06
Add some explanation of MPISharedArray type to the docs

johnomotani:


johnomotani:
@LucasMontoya4 do these new docs make the `MPISharedArray` any clearer? https://mabarnes.github.io/moment_kinetics/previews/PR246/developing/#Array-types

I'm not sure what I've written so far adds much, but I'm blanking at the moment on what else to put. It'd help if someone would ask questions to help decide what to add!

LucasMontoya4:
I think it makes more sense now. So the MPISharedArray exists in shared memory, but so will the Array when you run using mpi, and so they behave the same way?

johnomotani:
Julia doesn't actually know that our arrays are shared-memory or not (which is partly why we have to do all the synchronization and checking by hand). We use an MPI function to allocate the shared memory, then create a Julia `Array` that accesses that bit of memory, in this function https://github.com/mabarnes/moment_kinetics/blob/db370d14451aff4ab167e823fb774229923d0c8b/moment_kinetics/src/communication.jl#L550-L669

Normally, when debugging is not activated, `MPISharedArray` is just an alias for `Array`, defined here
https://github.com/mabarnes/moment_kinetics/blob/db370d14451aff4ab167e823fb774229923d0c8b/moment_kinetics/src/communication.jl#L546-L548
(I think I might write an actual docstring for that now...). When debugging is not activated, that macro expands out to give just
```julia
const MPISharedArray = Array
```
but if we hard-coded `Array` into our struct definitions, etc., then when we do activate debugging and want to replace `Array` with `DebugMPISharedArray`, we wouldn't be able to because `DebugMPISharedArray` isn't an `Array`. Julia only allows types to inherit from 'abstract types' not 'concrete types'. `Array` and `DebugMPISharedArray` both have to be concrete types (you can actually create instances of them), and they both inherit from `AbstractArray` (an abstract type, which effectively declares that a thing acts like an array, so generic functions know they can work with it). We don't want to put `AbstractArray` in our struct definitions because it's not a concrete type, so structs defined like that would not be type-stable.

LucasMontoya4:
That makes so much sense now, thank you! I didn't mean for you to write such a detailed comment, but I'm sure that that will help others in future as well...

PR #249, opened 2024-09-11
Correct 1 d matrix assembly

mrhardman:
Changes to correct the 1D matrix assembly functions. Documentation, simplification. Boundary conditions are now clearly only imposed when solving and ODE. Tests of this functionality could be added if desired. Demonstrated here so far is that differentiation does not require boundary conditions.

### DIFF
--- a/moment_kinetics/src/gauss_legendre.jl
+++ b/moment_kinetics/src/gauss_legendre.jl
@@ -120,17 +120,33 @@ function setup_gausslegendre_pseudospectral(coord; collision_operator_dim=true)
     L_matrix_with_bc = allocate_float(coord.n,coord.n)
 
     setup_global_weak_form_matrix!(mass_matrix, lobatto, radau, coord, "M")
+    mass_matrix_lu = lu(sparse(mass_matrix))
     setup_global_weak_form_matrix!(K_matrix, lobatto, radau, coord, "K_with_BC_terms")
     setup_global_weak_form_matrix!(L_matrix, lobatto, radau, coord, "L_with_BC_terms")
-    dirichlet_bc = (coord.bc == "zero") # and further options in future
+    dirichlet_bc = (coord.bc in ["zero", "constant"]) # and further options in future
     periodic_bc = (coord.bc == "periodic")
-    setup_global_weak_form_matrix!(L_matrix_with_bc, lobatto, radau, coord, "L", dirichlet_bc=dirichlet_bc, periodic_bc=periodic_bc)
-    mass_matrix_lu = lu(sparse(mass_matrix))
+    if dirichlet_bc || periodic_bc
+        setup_global_weak_form_matrix!(L_matrix_with_bc, lobatto, radau, coord, "L", dirichlet_bc=dirichlet_bc, periodic_bc=periodic_bc)
+    else
+        # Fill matrix with invertible identity matrix to avoid singular LU decomposition errors
+        # when no valid boundary condition supplied
+        identity_matrix!(L_matrix_with_bc, coord.n)
+        println("WARNING: L_matrix_with_bc replaced with identiy: 1D ODE solver requires boundary conditions")
+    end    
     L_matrix_lu = lu(sparse(L_matrix_with_bc))
+
     Qmat = allocate_float(coord.ngrid,coord.ngrid)
 
     return gausslegendre_info(lobatto,radau,mass_matrix,sparse(S_matrix),sparse(K_matrix),sparse(L_matrix),sparse(L_matrix_with_bc),
                               mass_matrix_lu,L_matrix_lu,Qmat)
+end
+
+function identity_matrix!(I,n)
+    @. I[:,:] = 0.0
+    for i in 1:n
+       I[i,i] = 1.0
+    end
+    return nothing    
 end
 
 function setup_gausslegendre_pseudospectral_lobatto(coord; collision_operator_dim=true)
--- a/moment_kinetics/src/gauss_legendre.jl
+++ b/moment_kinetics/src/gauss_legendre.jl
@@ -94,8 +94,12 @@ struct gausslegendre_info{TSparse, TLU} <: weak_discretization_info
     K_matrix::TSparse
     # global (1D) weak Laplacian derivative matrix
     L_matrix::TSparse
-    # global (1D) LU object
+    # global (1D) weak Laplacian derivative matrix with boundary conditions
+    L_matrix_with_bc::TSparse
+    # mass matrix global (1D) LU object
     mass_matrix_lu::TLU
+    # Laplacian global (1D) LU object
+    L_matrix_lu::TLU
     # dummy matrix for local operators
     Qmat::Array{mk_float,2}
 end
@@ -113,14 +117,20 @@ function setup_gausslegendre_pseudospectral(coord; collision_operator_dim=true)
     mass_matrix = allocate_float(coord.n,coord.n)
     K_matrix = allocate_float(coord.n,coord.n)
     L_matrix = allocate_float(coord.n,coord.n)
+    L_matrix_with_bc = allocate_float(coord.n,coord.n)
 
     setup_global_weak_form_matrix!(mass_matrix, lobatto, radau, coord, "M")
     setup_global_weak_form_matrix!(K_matrix, lobatto, radau, coord, "K_with_BC_terms")
     setup_global_weak_form_matrix!(L_matrix, lobatto, radau, coord, "L_with_BC_terms")
+    dirichlet_bc = (coord.bc == "zero") # and further options in future
+    periodic_bc = (coord.bc == "periodic")
+    setup_global_weak_form_matrix!(L_matrix_with_bc, lobatto, radau, coord, "L", dirichlet_bc=dirichlet_bc, periodic_bc=periodic_bc)
     mass_matrix_lu = lu(sparse(mass_matrix))
+    L_matrix_lu = lu(sparse(L_matrix_with_bc))
     Qmat = allocate_float(coord.ngrid,coord.ngrid)
 
-    return gausslegendre_info(lobatto,radau,mass_matrix,sparse(S_matrix),sparse(K_matrix),sparse(L_matrix),mass_matrix_lu,Qmat)
+    return gausslegendre_info(lobatto,radau,mass_matrix,sparse(S_matrix),sparse(K_matrix),sparse(L_matrix),sparse(L_matrix_with_bc),
+                              mass_matrix_lu,L_matrix_lu,Qmat)
 end
 
 function setup_gausslegendre_pseudospectral_lobatto(coord; collision_operator_dim=true)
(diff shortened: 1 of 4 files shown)


mrhardman:
@johnomotani Now I have checked in "test_scripts/GaussLobattoLegendre_test.jl" that the boundary condition options behave as expected by solving the 1D ODEs. I haven't provided a wrapper function to do this in a neat way, as I presume that the application will determine how general this function should be. I am happy to provide one if you have requirements, otherwise, I would ask for you input on how to put the additional 1D ODE tests in "test_scripts/GaussLobattoLegendre_test.jl" into the calculus tests for posterity. I think this could close some of my review comments to your PR #247.

mrhardman:
@johnomotani Unclear to me as to why calculus tests fail on ubuntu platform as they all pass locally for me.

johnomotani:
It looks like this PR actually changes the `K_matrix` and `mass_matrix` that are used for second derivatives, so I was confused how the tests were passing before. I think it might have been a bit of a fluke, because if we add a non-zero phase to the sin() in the function that's being tested, the gausslegendre tests with periodic bc fail without this PR and pass with it. I think we should add that phase to the tests as standard to make sure we're not missing something. @mrhardman I can add and push that if you like?

I'm not sure that's the end of the story, because even with this PR the output of `second_derivative!()` with periodic bc is not exactly periodic. I'm worried that if we ever used that, it would cause numerical instability. So I think either we need to fix periodic bc fully (maybe Patrick's email helps??) or just error out on weak-form second derivatives with periodic bc. I'd rather fix it, as long as it doesn't end up being too much work.

mrhardman:
> It looks like this PR actually changes the `K_matrix` and `mass_matrix` that are used for second derivatives, so I was confused how the tests were passing before. I think it might have been a bit of a fluke, because if we add a non-zero phase to the sin() in the function that's being tested, the gausslegendre tests with periodic bc fail without this PR and pass with it. I think we should add that phase to the tests as standard to make sure we're not missing something. @mrhardman I can add and push that if you like?

Sure, add extra tests with a phase if you think it is important. I am happy to add tests demonstrating the 1D ODE solve that you can check out in test_scripts/GaussLobattoLegendre_tests.jl. As far as I understood, all I changed here was reverting to a canonical assembly without factors of 2 on boundary rows. Those factors were originally there because I used some routines to form the RHS from the Chebyshev pseudospectral module, which average over boundary points, rather than just summing them. Happy that more complex tests pass on this PR!

> I'm not sure that's the end of the story, because even with this PR the output of `second_derivative!()` with periodic bc is not exactly periodic. I'm worried that if we ever used that, it would cause numerical instability. So I think either we need to fix periodic bc fully (maybe Patrick's email helps??) or just error out on weak-form second derivatives with periodic bc. I'd rather fix it, as long as it doesn't end up being too much work.

I haven't imposed any boundary conditions on results from differentiation, and I am quite sure that we should not impose boundary conditions in the mass matrix itself (it should be diagonally dominated like the identity). I am happy to impose boundary conditions in stiffness matrices if we are solving an ODE: I thought that Patrick's answer was giving options for improving boundary condition imposition in the ODE case. What level of error are we talking about? Is it enough to just copy the result from d2f[1] to d2f[end]?

johnomotani:
Yes, the Chebyshev (and probably finite difference) versions just average the first/last points for periodic bc.

The issue for the weak-form implementation is that for periodic bc the mass matrix does have to know about them - otherwise the result of `x = M^-1.b` does not have `x[1] = x[end]`.

As I understand Patrick's email, the point of 'Option A' is that the 'proper' thing to do would be not to have `x[end]` included in the numerical state at all. If you do that, then the first row of the mass matrix contains contributions both from the 'first' element and from the 'last' element, because the Lagrange polynomial corresponding to the first grid point extends over both of those elements, and similarly for the stiffness matrix.
If we insist on keeping `x[end]` on the grid, we can get the same effect by constructing the first row of the mass-matrix and stiffness-matrix in the same way, but zero out the last row of the stiffness matrix (so that the last element of the RHS vector in the `mass_matrix.x = b` solve is zero), and making the last row of the mass matrix (1 0 0 .... 0 0 -1), so that the last row of the linear system is the equation `x[1] - x[end] = 0`. I've implemented that, and it passes the stricter tests, and has x[1]=x[end], so it may have been a correct thing to do. Will push in a minute.

johnomotani:
Commenting to make my last commit (d971dc8) visible:
Ensure second derivatives give exactly periodic results without even rounding errors, as these might drift and accumulate
into a non-negligible error after many timesteps.

I'm not sure if this is really important, but worry that it might be. As we have never done second derivatives in periodic coordinates so far, this change (or the lack of it) does not affect any results.

mrhardman:
> Yes, the Chebyshev (and probably finite difference) versions just average the first/last points for periodic bc.
> 
> The issue for the weak-form implementation is that for periodic bc the mass matrix does have to know about them - otherwise the result of `x = M^-1.b` does not have `x[1] = x[end]`.
> 
> As I understand Patrick's email, the point of 'Option A' is that the 'proper' thing to do would be not to have `x[end]` included in the numerical state at all. If you do that, then the first row of the mass matrix contains contributions both from the 'first' element and from the 'last' element, because the Lagrange polynomial corresponding to the first grid point extends over both of those elements, and similarly for the stiffness matrix. If we insist on keeping `x[end]` on the grid, we can get the same effect by constructing the first row of the mass-matrix and stiffness-matrix in the same way, but zero out the last row of the stiffness matrix (so that the last element of the RHS vector in the `mass_matrix.x = b` solve is zero), and making the last row of the mass matrix (1 0 0 .... 0 0 -1), so that the last row of the linear system is the equation `x[1] - x[end] = 0`. I've implemented that, and it passes the stricter tests, and has x[1]=x[end], so it may have been a correct thing to do. Will push in a minute.

I am strongly against this solution, which I believe is wrong. Can we please revert this commit?

We need clarification from Patrick as to whether boundary conditions like this can be imposed on differentiation. I am sure that we cannot impose boundary conditions in the matrix unless we are solving an ODE.

moment_kinetics/src/gauss_legendre.jl
```diff
@@ -114,13 +120,34 @@ function setup_gausslegendre_pseudospectral(coord; collision_operator_dim=true,
...
-    setup_global_weak_form_matrix!(mass_matrix, lobatto, radau, coord, "M"; dirichlet_bc=dirichlet_bc)
-    setup_global_weak_form_matrix!(K_matrix, lobatto, radau, coord, "K_with_BC_terms"; dirichlet_bc=dirichlet_bc)
-    setup_global_weak_form_matrix!(L_matrix, lobatto, radau, coord, "L_with_BC_terms"; dirichlet_bc=dirichlet_bc)
+    dirichlet_bc = (coord.bc in ["zero", "constant"]) # and further options in future
+    periodic_bc = (coord.bc == "periodic")
+    setup_global_weak_form_matrix!(mass_matrix, lobatto, radau, coord, "M"; periodic_bc=periodic_bc)
+    setup_global_weak_form_matrix!(K_matrix, lobatto, radau, coord, "K_with_BC_terms"; periodic_bc=periodic_bc)
+    setup_global_weak_form_matrix!(L_matrix, lobatto, radau, coord, "L_with_BC_terms")
     mass_matrix_lu = lu(sparse(mass_matrix))
+    if dirichlet_bc #|| periodic_bc
```
mrhardman, line 129: || periodic_bc should not be commented out here.

moment_kinetics/src/gauss_legendre.jl
```diff
@@ -906,6 +913,26 @@ function setup_global_weak_form_matrix!(QQ_global::Array{mk_float,2},
...
+        j = coord.nelement_local
+        get_QQ_local!(QQ_j,j,lobatto,radau,coord,option)
+        iminl = imin[j] - (coord.nelement_local > 1)
+        imaxl = imax[j]
+        QQ_global[1,iminl:imaxl] .+= QQ_j[end,:]
+        # Enforce continuity at the periodic boundary
+        # All-zero row in RHS matrix sets last element of `b` vector in
+        # `mass_matrix.x = b` to zero
+        QQ_global[end,:] .= 0.0
+        if option == "M"
```
mrhardman, line 930: This if statement should not be here: if it is then we cannot use the periodic boundary conditions for solving ODEs.

moment_kinetics/src/gauss_legendre.jl
```diff
@@ -906,6 +913,26 @@ function setup_global_weak_form_matrix!(QQ_global::Array{mk_float,2},
...
             QQ_global[end,end] = 1.0
         end
+        # requires RHS vector b[1],b[end] = boundary values
+    end
+    if periodic_bc
+        # Make periodic boundary condition by modifying elements of matrix for duplicate point
+        # add assembly contribution to lower endpoint from upper endpoint
+        j = coord.nelement_local
+        get_QQ_local!(QQ_j,j,lobatto,radau,coord,option)
+        iminl = imin[j] - (coord.nelement_local > 1)
```
mrhardman, line 923: This line looks illegal. Can we use another line instead of this Julia feature (where presumably Bool => int?)

moment_kinetics/src/gauss_legendre.jl
```diff
@@ -906,6 +913,26 @@ function setup_global_weak_form_matrix!(QQ_global::Array{mk_float,2},
...
+        imaxl = imax[j]
+        QQ_global[1,iminl:imaxl] .+= QQ_j[end,:]
```
mrhardman, line 925: This looks wrong to me, we should only add on to the points which overlap between elements, not the whole row. My previous line was correct.

moment_kinetics/src/gauss_legendre.jl
mrhardman: Or an explicit type conversion to make it clear what is going on?

moment_kinetics/src/gauss_legendre.jl
mrhardman: I think I understand that the intention here is for this whole block to handle periodic BC for both the K and M matrices. However, there are two separate ways to use these matrices. If we want to solve an ODE, then the role of K and M is reversed, meaning that we need an M that is just multiplied (not LU decomposed), and a K that has BCs. This needs modification before it can go forward -- it is not general enough to be used.

mrhardman:
The last commit tries to make the ODE solver in "test_scripts/GaussLobattoLegendre_test.jl" work by putting two boundary conditions on `L_matrix_with_bc`. I think it would be desirable to put the ODE solver tests in to calculus_tests.jl. The logic in the gausslegendre.jl setup function is now a bit messy, and perhaps could do with cleaning up.

moment_kinetics/src/gauss_legendre.jl
johnomotani: I was thinking about that a bit - I think we might be able to treat K and M the same. Since we know that in the periodic case that `x[1] == x[end]`, then if the last row of the matrix is (1 0 0 ... 0 0 -1) then the last entry of `K.x` or `M.x` would still be zero, without having to zero out the whole of the last row. I haven't tested yet, but if that works it would mean we could use M and K either way around without any more special handling.

moment_kinetics/src/gauss_legendre.jl
johnomotani: Whatever you like, I don't mind.

moment_kinetics/src/gauss_legendre.jl
johnomotani: I've checked that this suggestion does work, so it's not necessary to distinguish between 'lhs' and 'rhs' matrices - I've just pushed a commit returning the 2 flags to one `periodic_bc`.

mrhardman: Nice! I think in practice one might still need to set b[end] = 0 exactly (b from K.x = b), but I am happy to leave the code as it is for now. Thanks for taking look.

mrhardman:
Final question: what automatic tests and functionality do we want from the 1D ODE solvers?

johnomotani:
> Final question: what automatic tests and functionality do we want from the 1D ODE solvers?

That is a good question. In general, anything that works should be tested so we don't break it, but I'm not using any ODE solvers at the moment, so happy to leave it to you.

I'm going to leave any tidying up and new tests for a future PR - the original PR #247 is about refactoring the input, not ODE solves and for that PR it's enough that the existing tests now pass.

PR #250, opened 2024-09-11
Multiple source functionality

LucasMontoya4:
Can now add multiple sources to the input files, where sources are numbered, so
```
[ion_source_1]
...

[ion_source_2]
...

[neutral_source_1]
...
```

All input files for tests and example input files have been updated, and plots can now be made for each source individually.


johnomotani:
Thanks @LucasMontoya4! Eventually we should add a test case that uses multiple sources, but that can wait until you've had time to work with them a bit and pick a favourite :slightly_smiling_face: I'll make an issue to remind us.

LucasMontoya4:
Thanks @johnomotani! I'm sorry about those typos - I hope they weren't too pesky to find...

PR #259, opened 2024-09-18
Remove mpi4py setup in parallel CI jobs

johnomotani:
Don't think it should cause a problem, but is unneeded now that we use OpenMPI from the Julia package manager for the CI jobs, and maybe there is some conflict that is causing intermittent failures in the parallel tests.

### DIFF
--- a/.github/workflows/debug_checks.yml
+++ b/.github/workflows/debug_checks.yml
@@ -15,9 +15,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - uses: mpi4py/setup-mpi@v1
-        with:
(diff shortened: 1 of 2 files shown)


johnomotani:
This change seems to significantly speed up the debug checks (which do run in parallel) and parallel tests. I don't understand why, as apart from not installing OpenMPI on the system (which doesn't take that long) this PR should not change anything - we should be using OpenMPI from the Julia package manager before and after. I guess maybe either installing OpenMPI on the system, or something else the mpi4py setup did, caused some sort of conflict. Or maybe this was just a fluke and we got a super fast server for the tests when they ran for this PR...

PR #265, opened 2024-09-26
Fix "wall_exp_decay" source for distributed MPI

johnomotani:
When distributed-memory MPI parallelism is used, `coord.grid` only has the positions of points in the local grid, not the global grid, so it doesn't necessarily contain the end points of the global grid. We have to use `coord.L` to get the locations of the walls instead.


johnomotani:
@LucasMontoya4 sorry for missing this in the review on #250!

LucasMontoya4:
Nono I should be sorry for missing that - and thank you for fixing it! I’ll remember that for future.

PR #269, opened 2024-10-01
Add input parameter zero_electric_field to permit testing of scaling …

### LINKED ISSUE #268: 1D2V simulations with electric field switch
To support Mantas Abazorius's work, it would be useful to implement a switch to turn off `Ez` so that the DKE solved for is 

```math
\frac{\partial F}{\partial t} + v_{\|}\frac{\partial F}{\partial z} = C[F,F]
```

@johnomotani @LucasMontoya4 Where should I implement the switch? Do you agree that this could be an interesting feature?

mrhardman:
…of F with vpa near wall without acceleration due to the sheath entrance.

Feature to support thesis work by Mantas Abazorius.

Suggestions and comments welcome. Closes #268.

### DIFF
--- a/moment_kinetics/src/em_fields.jl
+++ b/moment_kinetics/src/em_fields.jl
@@ -156,6 +156,15 @@ function update_phi!(fields, fvec, vperp, z, r, composition, collisions, moments
                     z_spectral,z)
         end
     end
+ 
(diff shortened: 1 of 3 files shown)


mrhardman:
Discussion in #268 shows that this PR is not necessary for required runs by @MantasAbazorius.

PR #270, opened 2024-10-01
Correct units for pressures to reflect normalisation used in the code.

### LINKED ISSUE #136: Factors of 2 in normalisation of pressure and definition of the thermal speed.
In the initial development of the 1D1V moment_kinetics models, the (parallel) pressure `ppar` was defined such that the thermal speed `vth` was calculated in the code like

`vth = sqrt(2*ppar/density)`

with `density` the species density. At the same time, the velocities were normalised to a reference speed `cref = sqrt(2 Tref /mref)`. Whilst self-consistent for a certain normalisation of the pressure, the above convention leads to a pressure that is 1/2 of the conventionally normalised pressure. Typically, we would normalise the pressure to `nref Tref = (1/2) mref nref cref^2`, where `nref` is a reference density. In this normalisation

`vth = sqrt(ppar/density)`

The branch https://github.com/mabarnes/moment_kinetics/tree/radial-vperp-standard-DKE-Julia-1.7.2-mpi currently has all pressures defined in the latter convention, where is the master branch currently uses the former convention. Developers beware!

N.B. This discussion assumes that all masses are unity.

To Do: agree to a standard normalisation convention that are natural for both 1V and 2V simulations (where pperp and ppar are present).

mrhardman:
Units for the output binary file appear to be incorrect for the pressure. Here I have corrected the units to what I understand the normalisations are.

The old normalisation for p was 
```math
  p_{ref} = n_{ref} T_{ref}
```
the corrected normalisation is 
```math
 p_{ref} = n_{ref} m_{ref} c_{ref}^2
```
with $`c_{ref} = \sqrt{2 T_{ref}/m_{ref}}`$.

@johnomotani We should also check that the normalisation reported for the heat flux is correctly documented, since the units there also reference $`T_{ref}`$ and not $`m_{ref}c_{ref}^2`$.

May close #136.

### DIFF
--- a/moment_kinetics/src/file_io.jl
+++ b/moment_kinetics/src/file_io.jl
@@ -1563,13 +1563,13 @@ function define_dynamic_electron_moment_variables!(fid, r::coordinate, z::coordi
     io_electron_ppar = create_dynamic_variable!(dynamic, "electron_parallel_pressure", mk_float, z, r;
                                        parallel_io=parallel_io,
                                        description="electron species parallel pressure",
-                                       units="n_ref*T_ref")
+                                       units="n_ref*m_ref*c_ref^2")
     if write_error_diagnostics
         io_electron_ppar_loworder =
             create_dynamic_variable!(dynamic, "electron_parallel_pressure_loworder", mk_float,
                                      z, r; parallel_io=parallel_io,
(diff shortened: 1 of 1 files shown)


johnomotani (approved):
I think this is correct. I've applied a similar update to the units for heat fluxes.

Issue #275, opened 2024-10-08
Precompile bugs?


mrhardman:
When I try to precompile on a machine which is usually reliable, I now see this output

```
┌ CompositionsBase → CompositionsBaseInverseFunctionsExt
│  [pid 2781818] waiting for IO to finish:
│   Handle type        uv_handle_t->data
│   fs_event           0xcce120->0x7f36fd34cdf0
│   timer              0xe0f4f0->0x7f36fd34ce20
│  This means that a package has started a background task or event source that has not finished running. For precompilation to complete successfully, the event source needs to be closed explicitly. See the developer documentation on fixing precompilation hangs for more help.
└
┌ LoweredCodeUtils
│  WARNING: using IR.SSAValue in module LoweredCodeUtils conflicts with an existing identifier.
│  WARNING: using IR.SlotNumber in module LoweredCodeUtils conflicts with an existing identifier.
└
```

Is this caused by changes related to #273?

I have checked that the branch here https://github.com/mabarnes/moment_kinetics/commits/electric_field_switch/ without the timer changes does not have the first error message (the `LoweredCodeUtils`  error is present). In this older branch, I can still use 
```
julia --project -O3 precompile.jl
```
to make `moment_kinetics.so` and run
```
mpirun -n 2 julia --project -O3 -Jmoment_kinetics.so -e 'include("moment_kinetics/test/runtests.jl")'
```
successfully. However, in the latest master, with the manual install instructions I get the following error when I run the above command
```
--------------------------------------------------------------------------
It looks like orte_init failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during orte_init; some of which are due to configuration or
environment problems.  This failure appears to be an internal failure;
here's some additional information (which may only be relevant to an
Open MPI developer):

  getting local rank failed
  --> Returned value No permission (-17) instead of ORTE_SUCCESS
--------------------------------------------------------------------------
--------------------------------------------------------------------------
It looks like orte_init failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during orte_init; some of which are due to configuration or
environment problems.  This failure appears to be an internal failure;
here's some additional information (which may only be relevant to an
Open MPI developer):

  orte_ess_init failed
  --> Returned value No permission (-17) instead of ORTE_SUCCESS
--------------------------------------------------------------------------
--------------------------------------------------------------------------
It looks like MPI_INIT failed for some reason; your parallel process is
[...]

PR #278, opened 2024-10-09
Option to use Julia-provided MPI in machine setup

johnomotani:
Occasionally it can be useful to use the Julia-provided MPI instead of the system MPI, for example to use Julia-provided binaries for external libraries that depend on MPI (e.g. the MUMPS library that provides MPI-parallelised LU). This PR adds an option to choose the Julia-provided MPI in the `machine_setup.sh` script.

When using the Julia-provided MPI, you have to use the corresponding `mpiexec` command. MPI.jl provides a function to generate a link to this command - the machine setup script generates this link in the project directory, so you would run an MPI job like, for example,
```bash
./mpiexecjl -np 8 julia --project -O3 --check-bounds=no run_moment_kinetics examples/nonlinear-sound-wave/nonlinear-sound-wave_cheb_split3.toml
```


mrhardman:
Thanks for suggesting this feature? Perhaps making a comment in the manual setup to link to https://juliaparallel.org/MPI.jl/stable/usage/ and https://juliaparallel.org/MPI.jl/stable/configuration/#configure_jll_binary could also be helpful? 

Have you sucessfully used the  `mpiexecjl` command with a `-Jmoment_kinetics.so` system image? I am trying this now to see if this solves my issue #275. However, on the first attempt with
```
julia --project -O3 precompile.jl
.julia/bin/mpiexecjl --project=./ -n 2 julia --project -O3 -Jmoment_kinetics.so -e 'include("moment_kinetics/test/runtests.jl")'
```
no jobs start, seeming like a hang. I wonder if that is because `mpiexecjl` only supports `-O0`.

PR #279, opened 2024-10-09
Fix machine setup for HDF5 on macOS

johnomotani:
Need to us a different library extension (".dylib" instead of ".so") on macOS.


johnomotani:
Don't know why the parallel macOS tests are failing, but that can't be caused by this PR, because this only changes a setup script that isn't used in the CI jobs, so going to merge anyway.

PR #282, opened 2024-10-14
Update CI jobs and documentation build to Julia-1.11

johnomotani:
I got these updates ready, but I'm finding there are a few glitches with Julia-1.11.0 (it seems to crash when I try to use the `pkg>` interface from the REPL). There seems to be a fix for that problem which will be released in 1.11.1, so probably best to hold off on merging this at least until 1.11.1 is released.


johnomotani:
CI tests pass now using Julia-1.11.3. I've run some simulations with 1.11.3 to test #316 as well, which worked without problems, so I think we can start using 1.11.x now.

PR #285, opened 2024-10-15
Suggestions for online documentation of manual setup

mrhardman:
To try address the problems encountered in #275, here I have tried to update the online documentation for the manual install, to reflect the current state of the machine setup scripts.

### DIFF
--- a/docs/src/manual_setup.md
+++ b/docs/src/manual_setup.md
@@ -63,6 +63,36 @@ file is named something other than `libmpi.so`, you might have to pass some
 keyword arguments to `use_system_binary()` - see
 <https://juliaparallel.org/MPI.jl/stable/reference/mpipreferences/#MPIPreferences.use_system_binary>.
 
+## Using the native Julia MPI
+
+As an alternative, you can also use the MPI that is shipped with Julia.
+```julia
+julia> using MPI; MPI.install_mpiexecjl(force=true)
+julia> using MPIPreferences; MPIPreferences.use_jll_binary()
+```
+The executable for the Julia MPI can be called from the root folder of the project with
+```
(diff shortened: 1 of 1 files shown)


johnomotani:
We could maybe simplify a bit for the dependency packages. Things like `StatsBase` are already dependencies of `moment_kinetics`, so we could import them via `moment_kinetics` to avoid needing to add them to the 'top-level' requirements. It's slightly uglier in the test/precompile scripts, but probably worth it in terms of making setup simpler. We can add a paragraph in the developer docs to explain why we handle some dependencies that way. If that sounds like a good idea, I can push changes to this PR (since those changes would also need changes to the docs, it seems reasonable to do it this way).

Are you sure that `Test` needs to be added explicitly? I think it's part of the stdlib that's bundled with Julia, so it'd be a bit surprising if we need to explicitly add it.

mrhardman:
I am sure that `Test` needs to be added explicitly. Without it, I found faults with the loading of extensions of dependency modules. I think you can see this in this post https://github.com/mabarnes/moment_kinetics/issues/275#issuecomment-2413104854. Although in that case I had pinned enough modules to make things work, it turned out that I could get runs with no errors and precompilation if I just used the version independent recipe in https://github.com/mabarnes/moment_kinetics/issues/275#issuecomment-2413943529.

As to loading modules like `StatsBase` and `SpecialFunctions` via moment_kinetics for `moment_kinetics/test/*.jl`, I am in favour of this solution, as it avoids having to load modules just for the test routines that we would like to `just work` for new users. The `test_scripts/*.jl` files do tend to need modules that we should not load (like `Plots`) -- but since all these files are my precursor test experiments (and Fokker-Planck error plot generators), I am happy to keep loading the modules for those manually.

PR #302, opened 2025-01-11
Better handling of wall boundary conditions in implicit electron timesteps

johnomotani:
* The $v_\mathrm{cut}$ specifying the beginning of the tail of the electron distribution function that is lost to the wall is recalculated at each Newton iteration in the electron implicit timestep. However, the constraints are not enforced as they would be when imposing the full boundary condition (as 'hard forcing' of constraints interferes with the convergence of the Newton-Krylov iteration).
  * This seems to significantly enhance the robustness of the electron implicit timesteps.
* Adds the contribution of the wall boundary condition in the function that calculates the electron Jacobian matrix.
  * Not thoroughly tested yet, but initial indications are that this addition helps a little in increasing the electron timestep and/or reducing the number of iterations, but on the level of ~10-20%.


johnomotani:
Parallel tests are failing in the kinetic electron test with ADI preconditioner, on 2 cores. I'm not sure why, and I can't reproduce the error on my laptop. The test failure seems to have gone away in #303, so I'm going to ignore it and hope it was not a real problem.

PR #306, opened 2025-01-14
Final transfer version

LucasMontoya4:
Transfer multi source function and PI controller, and also functionality to turn off kinetic distribution function evolution and use a fluid closure for the ions instead. Note that this still only speeds up code by factor of around 5, as a lot of functions operating on f are embedded in essential code for the fluid model. Hence, best way to speed things up is to make ngrid and nelement 1 for vpar.


examples/multi-sources/krook_closure_temp_PI_density_PI_source.toml
mrhardman, line 1: If you want to keep a revisioned copy of these files you could just make a new directory to store them in. It looks like only examples/* is checked for *.toml files in the workflow run. https://github.com/mabarnes/moment_kinetics/blob/master/.github/workflows/examples.yml.

LucasMontoya4: I guess that's true, but I'm dumping them all in the publication_inputs folder anyway, so we can keep the examples folder clean.

johnomotani:
I found the hang when running the PI controller tests in parallel. It was in loading the data after the run. `get_run_info_no_setup()` passes a flag `ignore_MPI=true` to `mk_input()` to tell it to skip MPI-ish things because it's being called in serial (even if there are multiple MPI ranks present, like in the parallel tests). That flag needed to be passed through to `setup_external_sources!()` to avoid allocating shared-memory arrays or doing MPI operation when `ignore_MPI=true`. That was a bug I made when I wrote the initial PI controller, not one from this PR - just shows that we do need tests for everything!

johnomotani (approved):
Looks good to me. Nice work @LucasMontoya4! I'll just add a couple of little tidy-up changes, then merge this once the tests pass.

examples/multi-sources/krook_closure_temp_PI_density_PI_source.toml
johnomotani: > It looks like only examples/* is checked for *.toml files in the workflow run.

BTW this is not true. The `walkdir("examples/")` function recursively runs over all the entries of all the subdirectories of `examples/` as well as the directory itself. So @LucasMontoya4's solution is the only possible one.

LucasMontoya4:
Thanks for doing this @johnomotani! I hope the code review didn't take too long...

johnomotani:
@LucasMontoya4 it was fine - I didn't check all your input files though!

PR #307, opened 2025-01-17
[WIP] Swap serial lu for MPI parallel qr using MPIQR.jl

jwscook:
Currently running on a `dev`'d MPIQR.jl with 3-argument `ldiv!` implemented amongst other things...


jwscook:
Right now `MPIQR.jl` is unregistered so you have to `pkg> add git@github.com:jwscook/MPIQR.jl.git`

johnomotani:
In case someone else wants to try this
```
add git@github.com:jwscook/MPIQR.jl.git
```
gave me a prompt asking for a private key to access Github, but just repeated the prompt when I gave it the key. Using instead
```
add https://github.com/jwscook/MPIQR.jl.git
```
just worked.

jwscook:
LU decomposition is able to exploit sparsity whereas QR decomposition is not. This means there's a huge performance gap between the two approaches. An example matrix has 2% non-zeros, hence a dense method is not going to win-out.

For reference, and thanks to @johnomotani:

```
LU solvers, running in serial:

sparse:
ldiv! 13.0ms
lu! 1.12s
lu 1.52s

dense:
ldiv! 102ms
lu 72.5s
```

PR #316, opened 2025-02-04
@inbounds in loop macros

johnomotani:
Add an `@inbounds` to the definition of `@loop_*` macros. This ensures that we skip bounds checking in most of the performance-critical parts of the code, without having to use `--check-bounds=no` (which is slightly broken in Julia-1.10 and more broken in Julia-1.11).

If there are performance-critical places that do not use `@loop_*` macros, then explicit `@inbounds` is required. One such place is `fokker_planck_calculus`, which has had `@inbounds` added in this PR.

Note bounds checking can be switched on for testing/debugging with `--check-bounds=yes`.

Replaces #315.

### DIFF
--- a/.github/workflows/parallel_test.yml
+++ b/.github/workflows/parallel_test.yml
@@ -21,15 +21,15 @@ jobs:
       - uses: julia-actions/cache@v2
       - run: |
           touch Project.toml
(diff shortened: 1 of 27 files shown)


johnomotani:
I have tested that without using `--check-bounds=no`, and on Julia-1.10.7 or Julia-1.11.3, this PR matches (or slightly exceeds), the performance of the code on Julia-1.10.7 with `--check-bounds=no`. I tested a 1D1V moment-kinetic ion, Boltzmann electron case, a 1D2V case with Fokker-Planck collisions and Boltzmann electrons, and a 1D1V case with kinetic electrons.

PR #318, opened 2025-02-10
Include @views to reduce allocations in electron functions

johnomotani:
Not sure why these lines cannot get to zero allocations, might be a 'bug' in the Julia compiler...


johnomotani:
Forgot what I was doing with this branch, I think the real changes were merged already. Closing and deleting branch.

PR #320, opened 2025-02-13
Simplify calculation of coefficients in moment-kinetic equation

johnomotani:
Saving the time derivatives of the moments as they are calculated for the time-advance of the moments, and then using those time derivatives for the coefficients in the moment-kinetic equation, simplifies the code, making fewer places that need to be updated if new
terms are added (e.g. if new terms are added to the RHS of the drift-kinetic equation, only the moment equations and the RHS of the moment-kinetic equation change, not the $\dot z$, $\dot w_\parallel$, $\dot F$ coefficients).

This change does not apply to electrons (only ions and neutrals). In the kinetic equation for the electron shape function, the mass ratio expansion in small $\sqrt{m_e/m_i}$ is done after substituting the moment equations into the kinetic equation, so it does not make sense to use $\partial p_{e\parallel}/\partial t$ as calculated for the evolution of $p_{e\parallel}$ in the kinetic equation.

Updates expected values in some tests: discretization errors are changed slightly by the rearrangement of the moment kinetic advection coefficients, and the tests are generally poorly resolved (in order to be fast), so are sensitive to this.

There is a question whether making this change is a good idea, with the concern that there may be some desirable cancellations between terms that would take place analytically but might not happen exactly numerically. Possibly the most concerning is the cancellation of the parallel electric field, but I think that one should be exact up to machine precision, as 
```julia
dupar_dt = dnupar_dt / 	n + ...
                 = (0.5 * Ez * n + ...) / n + ...
wpa_speed = (0.5 * Ez - dupar_dt + ...) / vth
```
So that we end up evaluating numerically `0.5 * Ez - 0.5 * Ez * n / n` - there will be some rounding error, but no discretization error (for other terms there might be a discretization error due to something like `d(n*upar)/dz - n*dupar/dz - upar*dn/dz`). For the corresponding lines of code, see:
https://github.com/mabarnes/moment_kinetics/blob/e8d95c56cece3c5124ca4c34daf2cc7a2cf74fdb/moment_kinetics/src/velocity_moments.jl#L962
https://github.com/mabarnes/moment_kinetics/blob/e8d95c56cece3c5124ca4c34daf2cc7a2cf74fdb/moment_kinetics/src/force_balance.jl#L31
https://github.com/mabarnes/moment_kinetics/blob/e8d95c56cece3c5124ca4c34daf2cc7a2cf74fdb/moment_kinetics/src/vpa_advection.jl#L410-L413


johnomotani:
@mrhardman ah, my comment was not correct (I'll edit to try and clarify). I was only thinking of adding new terms to the RHS of the kinetic equation, not potentially adding drift terms or other modifications to the free-streaming or acceleration. Potentially we could do something along the lines you're suggesting for drift terms - I think it's something to come back to when we're implementing the 2D version of moment-kinetics.

Yes, using arrays with dn/dt, etc. instead of substituting by hand using the moment equations, then evaluating the resulting expressions is what this PR does.

mrhardman:
> @mrhardman ah, my comment was not correct (I'll edit to try and clarify). I was only thinking of adding new terms to the RHS of the kinetic equation, not potentially adding drift terms or other modifications to the free-streaming or acceleration. Potentially we could do something along the lines you're suggesting for drift terms - I think it's something to come back to when we're implementing the 2D version of moment-kinetics.
> 
> Yes, using arrays with dn/dt, etc. instead of substituting by hand using the moment equations, then evaluating the resulting expressions is what this PR does.

Yes, in the best case scenario when modifying the advection velocities and accelerations we would have to modify both $`\dot{z}`$, $`\dot{w}_\|`$, $`\dot{w}_\perp`$, and $`\partial n/ \partial t`$,  $`\partial u_\|/ \partial t`$, and  $`\partial p/ \partial t`$. I think my suggestion was aimed at keeping the effort in those modifications to the theoretical minimum, happy for you to consider this at a later date if that is not the point of this PR.

moment_kinetics/src/time_advance.jl
```diff
@@ -3450,6 +3450,80 @@ with fvec_in an input and fvec_out the output
...
     end
 
+    # Start advance for moments
+    if advance.continuity
+        continuity_equation!(fvec_out.density, fvec_in, moments, composition, dt,
+                             z_spectral, collisions.reactions.ionization_frequency,
+                             external_source_settings.ion, num_diss_params)
+    end
+    if advance.force_balance
+        # fvec_out.upar is over-written in force_balance! and contains the particle flux
```
johnomotani, line 3460:
```suggestion
```

moment_kinetics/src/time_advance.jl
```diff
@@ -3450,6 +3450,80 @@ with fvec_in an input and fvec_out the output
...
+                         composition, external_source_settings.ion, num_diss_params)
+    end
+    if advance.neutral_continuity
+        neutral_continuity_equation!(fvec_out.density_neutral, fvec_in, moments,
+                                     composition, dt, z_spectral,
+                                     collisions.reactions.ionization_frequency,
+                                     external_source_settings.neutral, num_diss_params)
+    end
+    if advance.neutral_force_balance
+        # fvec_out.upar is over-written in force_balance! and contains the particle flux
```
johnomotani, line 3476:
```suggestion
```

PR #321, opened 2025-02-19
Format sub-sections in 'Magnetic Geometry' docs

johnomotani:
Formatting different levels of sub-sections makes the contents sidebar on the docs page look nicer.


johnomotani:
Test failures are due to some problem with NCDatasets.jl, which is failing to link correctly at the moment with MPI. Probably NCDatasets.jl will fix itself soon, so we can ignore the failures. Besides, this PR only updates the docs!

PR #322, opened 2025-02-19
Consistent 1D2V and 1D1V moment-kinetic equations

johnomotani:
Update to prepare for extending the moment-kinetic equations in the code from 1D1V to 1D2V (and in future nD2V).

* [x] Type up notes by @slnewton showing how to derive the 1D1V moment-kinetic equations as the $T_\perp \rightarrow 0$ limit of 1D2V moment-kinetic equations.
    * This process suggests some redefinitions of the 1D1V system that would make it more consistent with the 1D2V and therefore reduce the number of places in the code that need to have separate branches for 1D1V and 1D2V.
* [x] Simplify the conversion to dimensionless equations to be implemented in code.
    * Point for discussion - does anyone have suggestions or objections to the definitions I've proposed in the docs in this PR?
    * Aim in the proposal is to make the conversion of equations as simple as possible, and the conversion of dimensionless quantities back to SI units. For the current set of equations, the proposed definitions make the conversion 'put a hat on every non-dimensionless variable' - this I think won't still be true in 2D/3D with drifts included, but would like to stay as close as possible.
* [x] Update code, tests, and example inputs in line with the new definitions.
    * The tests all pass without generating any new expected output (except for the Braginskii electrons test, where results changed due to some slightly different choices of definitions a6472a4), the expected results are only scaled according to the factors in the table in the docs [on this page](https://mabarnes.github.io/moment_kinetics/previews/PR322/moment_kinetic_equations/#Conversion-to-old-dimensionless-equations) - a couple of 'hacks' were needed to exactly match old behaviour (ba1a711 and 0faf6fc), which I'll remove in a future PR.

This PR upgrades `util/update-input-file.jl` so that it (mostly) retains the structure of and comments in the original input file when updating it.


johnomotani:
The derivation and proposed definitions are on this updated page in the docs
https://mabarnes.github.io/moment_kinetics/previews/PR322/moment_kinetic_equations/#Moment-kinetic-equations

mrhardman:
Comments so far:

 - I am wondering what you will do in the source about naming of the evolved pressure variable `ppar`, when the evolved pressure in 2V is the total pressure -- will we use the `ppar` variable as the pressure, or will there be a new array pressure that contains both `ppar` and `pperp`? It looks like we will evolve total pressure, but will the code be refactored to make that clear?
 - Will the factors of  `pi^(1/2)` and `pi^(3/2)` be banished from the source in the normalisation of `pdf`? These factors are not in your normalisations for $F$.
 - I am not sure how useful it is to keep the notes on conversion between only and new conventions once the code refactoring is done. I understand why it is there now, but it should be put into some archive so that the then present definitions are easy to find.

mrhardman:
- Another point to consider in making 1D1V consistent with 1D2V is the application of the moment constraints. It would be good to use the same set of functions for both. Compare and contrast the routines https://github.com/mabarnes/moment_kinetics/blob/b3b2c86cbdc44016021cfd3c38f7447d938ada3a/moment_kinetics/src/fokker_planck.jl#L676-L721 and https://github.com/mabarnes/moment_kinetics/blob/b3b2c86cbdc44016021cfd3c38f7447d938ada3a/moment_kinetics/src/moment_constraints.jl#L29-L92.

Since by convention, `vperp.grid = 0` when `nvperp = 1`, we should be able to use the 2V matrix from the collision operator implementation in terms of "physical" moment integrals to also carry out the 2V and 1V moment constraints, with suitable modifications from the original context.

johnomotani:
Have now updated code for Boltzmann electron simulations (without FP collision operator). Appropriately converted tests pass without any changes to expected results, except multiplication by expected scaling factors.

Electron code, FP collisions and postprocessing code still to-do.

Developer warning: this PR still a draft because I expect to rebase the branch several more times, so any branches based on the current version here will likely also need to be rebased on the final version of this branch.

johnomotani:
I think this PR is ready now. The tests all pass without generating any new expected output (except for the Braginskii electrons test, where results changed due to some slightly different choices of definitions a6472a4), the expected results are only scaled according to the factors in the table in the docs [on this page](https://mabarnes.github.io/moment_kinetics/previews/PR322/moment_kinetic_equations/#Conversion-to-old-dimensionless-equations) - a couple of 'hacks' were needed to exactly match old behaviour (ba1a711 and 0faf6fc), which I'll remove in a future PR.

moment_kinetics/src/reference_parameters.jl
```diff
@@ -61,7 +61,12 @@ function setup_reference_parameters(input_dict, warn_unexpected::Bool)
...
 """
-Calculate normalized ion-ion collision frequency at reference parameters for Coulomb collisions.
+Calculate normalized ion-ion collision frequency for Coulomb collisions using reference
+density, mass and speed.
+Note this is not the thermal collision frequency evaluated with reference parameters, but
+rather the quantity that when multiplied by \$\\hat{n}s/v_{Ts}^3\$ gives the
+de-dimensionalised thermal collision frequency - the difference is a factor of
+\$2^{3/2}\$.
 
 Currently valid only for hydrogenic ions (Z=1)
```
mrhardman, line 71: Not directly related to the changes here, but since the docs are changed above, it is worth noting that I only envisage calculating the result here for hydrogen and multiplying and dividing by appropriate factors of Z (charge #) and A (mass) in the `fokker_planck.jl` to construct the correct prefactor for the operator. If we were to try to include the variation of the Coulomb logarithm, we would need to do more.

moment_kinetics/src/external_sources.jl
```diff
@@ -949,22 +1032,23 @@ Add external source term to the ion kinetic equation.
...
         end
     elseif source_type == "beam" || source_type == "beam-with-losses"
         @begin_s_r_z_region()
         source_vpa0 = ion_source.source_vpa0
         source_vperp0 = ion_source.source_vperp0
+        Maxwellian_prefactor = 1.0 / π^1.5
         if !(source_vpa0 > 1.0e-8)
             error("source_vpa0=$source_vpa0 < 1.0e-8")
         end
         if !(source_vperp0 > 1.0e-8)
```
mrhardman, line 1045: Why is this check here? Why can't I have a source with no vperp0?

moment_kinetics/src/external_sources.jl
```diff
@@ -949,22 +1032,23 @@ Add external source term to the ion kinetic equation.
...
             error("Unsupported combination in source_type=$(source_type) evolve_density=$(moments.evolve_density), "
-                  * "evolve_upar=$(moments.evolve_upar), evolve_ppar=$(moments.evolve_ppar)")
+                  * "evolve_upar=$(moments.evolve_upar), evolve_p=$(moments.evolve_p)")
         end
     elseif source_type == "beam" || source_type == "beam-with-losses"
         @begin_s_r_z_region()
         source_vpa0 = ion_source.source_vpa0
         source_vperp0 = ion_source.source_vperp0
+        Maxwellian_prefactor = 1.0 / π^1.5
         if !(source_vpa0 > 1.0e-8)
```
mrhardman, line 1042: Why is this check here? Surely vpa0 can be negative?

mrhardman (changes requested):
Thank you for this huge amount of work! Besides are few questions, which you can find in my comments, my main comments are 
 - I am happy with the refactoring of moment integration and the changes in Fokker-Planck related modules.
 - It is unclear to me if the moment integration structure you have here intends to support mass /=1 or not, and whether or not you intend to include the normalised mass ratio $`m/m_{ref}`$ in the definitions of pressure, heatflux, etc. If this feature is not intended to be available now, we will certainly need it for me to implement cross-species collisions, so agreeing how things should work is important before we undertake multi-species development. If you have decided already how things should be w.r.t. mass /= 1 species, please point me to an example.
 - The use of unicode `pi` is a problem for me in the usual editor that I use, along with any other unicode characters which are not found on a keyboard. Could we please avoid using greek character unicode in the source? Is it possible to remove unicode `pi`? Unicode in docstrings might be fine, but if I edit these files in the future I will replace unicode pi to `pi` so that I can read what I am doing.
 - We have a huge number of files in this repo, and there are 1109 files changes in this PR! I have thus only reviewed ones for which I am actively developing, i.e., relating to Fokker-Planck. Whilst it is nice to keep all examples and results in this repo, the number of files in this review suggests that some of those files should be stored in their own repositories.

moment_kinetics/src/fokker_planck_test.jl
```diff
@@ -210,7 +210,7 @@ Function computing \$ F_{\\rm Maxwellian} \$.
 function F_Maxwellian(dens::mk_float,upar::mk_float,vth::mk_float,
                         vpa,vperp,ivpa,ivperp)
     eta = eta_func(upar,vth,vpa,vperp,ivpa,ivperp)
-    fac = (dens/(vth^3))*exp(-eta^2)
+    fac = (dens/(vth^3)/π^1.5)*exp(-eta^2)
```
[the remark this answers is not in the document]
johnomotani: What editor are you using? I think you said notepad++, but on a fresh install (on Windows) that displays unicode fine for me (so does the built-in Notepad)
julia-code-notepad++
and apparently should have done since at least 2009 (https://superuser.com/questions/21135/how-can-i-edit-unicode-text-in-notepad).

johnomotani (commented):
Thanks for the review @mrhardman - see below.

moment_kinetics/src/external_sources.jl
mrhardman: Whoops, I should have checked further. I don't recall what is going on here, clearly.

moment_kinetics/test/fokker_planck_tests.jl
```diff
@@ -537,7 +539,7 @@ function runtests()
                         end
                     else
                         atol = 1.0e-4
-                        @test isapprox(dSdt, 2.543251178128757 ; atol=atol)
+                        @test isapprox(dSdt, 2.543251178128757 / pi^1.5 ; atol=atol)
```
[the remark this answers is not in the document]
mrhardman: OK, I will understand all this in detail when I have to merge master into #353 .

moment_kinetics/test/fokker_planck_tests.jl
```diff
@@ -495,18 +495,19 @@ function runtests()
                         end
                         @test isapprox(dSdt, rtol ; atol=atol)
                         delta_n = get_density(C_M_num, vpa, vperp)
-                        delta_upar = get_upar(C_M_num, vpa, vperp, dens)
-                        delta_ppar = msp*get_ppar(C_M_num, vpa, vperp, upar)
-                        delta_pperp = msp*get_pperp(C_M_num, vpa, vperp)
-                        delta_pressure = get_pressure(delta_ppar,delta_pperp)
+                        delta_upar = get_upar(C_M_num, dens, vpa, vperp, false)
+                        delta_pressure = msp*get_p(C_M_num, dens, upar, vpa, vperp, false, false)
```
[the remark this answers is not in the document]
mrhardman: I think previously I did have the facility to include non-unit masses in some of the moments, for tests. It might be that this was lost at some point. Either way, please do include me when you do consider mass /= 1, as those choices will affect my further development.

mrhardman:
@johnomotani I think you can go ahead, https://github.com/mabarnes/moment_kinetics/pull/322/commits/8ac82c62192192dd90cad2eda7ae4b2b5f4a6d1b addresses my main naming concern. We can discuss mass /= 1 again when nspecies > 1 comes up.

mrhardman (approved):

PR #325, opened 2025-02-24
Remove PyPlot from plots_post_processing to permit development withou…

mrhardman:
…t solving Python/Julia install issues.

This PR to remove PyPlot from `plots_post_processing`, to remove install issues when using  `plots_post_processing` to diagnose standard drift kinetic runs. I understand that @johnomotani no longer uses this package, so hopefully this change for convenience will not cause alarm.

### DIFF
--- a/plots_post_processing/plots_post_processing/src/plots_post_processing.jl
+++ b/plots_post_processing/plots_post_processing/src/plots_post_processing.jl
@@ -17,10 +17,6 @@ export get_coords_ngrid
 include("post_processing_input.jl")
 include("../../../makie_post_processing/makie_post_processing/src/shared_utils.jl")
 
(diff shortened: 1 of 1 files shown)


johnomotani:
Note that this removes the 'unnormalised distribution function' plot functionality from plots_post_processing - these are the functions that would plot a moment-kinetic distribution function as a function of unnormalised v_parallel instead of w_parallel.

I'm happy for this to go. The functionality is available in makie_post_processing, and having to use PyPlot.jl was very annoying, although matplotlib's pcolormesh is (or was when I last checked) significantly faster than the Julia equivalent (which is a slightly hacky workaround because the direct equivalent is not implemented in Makie.jl) when the plotting coordinates are 2d, rather than 1d arrays, so there could be a motivation to revive the code removed here at some point in the future.

PR #327, opened 2025-03-09
Temporarily disable NCDatasets in some CI tests

johnomotani:
CI tests on Github Actions are failing due to segfaults in NCDatasets calls. The parallel tests do not seem to be affected, don't know why - maybe because they compile a system image??

Hopefully the upstream packages will sort this out eventually. To keep an eye on when this happens, have kept an 'examples' test job that includes NCDatasets, which is expected to fail at the moment. When that job passes again, we should revert this commit.


johnomotani:
Not needed - see #328.

PR #334, opened 2025-03-19
Add option in timestepping section to print out density or temperatur…

LucasMontoya4:
…e at midpoint (electron steps not supported yet), which can be useful when gauging effectiveness of PI controllers - you can essentially 'see' the graphs of density and temperature motion without having to use makie_post_processing

### DIFF
--- a/moment_kinetics/src/time_advance.jl
+++ b/moment_kinetics/src/time_advance.jl
@@ -2025,6 +2025,14 @@ function  time_advance!(pdf, scratch, scratch_implicit, scratch_electron, t_para
                               rpad(string(t_params.moments_output_counter[]), 4), "  ",
                               "t = ", rpad(string(round(t_params.t[], sigdigits=6)), 7), "  ",
                               "nstep = ", rpad(string(t_params.step_counter[]), 7), "  ")
+                        if t_params.print_nT_live
+                            midpoint = Int64(round(size(moments.ion.dens)[1]/2))
+                            print("midpoint density: ", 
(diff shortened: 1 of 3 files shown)


moment_kinetics/src/time_advance.jl
```diff
@@ -2114,6 +2115,14 @@ function  time_advance!(pdf, scratch, scratch_implicit, scratch_electron, t_para
...
                                 "nstep = ", rpad(string(t_params.step_counter[]), 7), "  ",
                                 Dates.format(now(), dateformat"H:MM:SS"))
+                        if t_params.print_nT_live
+                            midpoint = Int64(round(size(moments.ion.dens)[1]/2))
+                            print("midpoint density: ", 
+                            rpad(string(round(moments.ion.dens[midpoint,1,1], sigdigits = 4)), 7))
+                            print("   midpoint temperature: ", 
+                            rpad(string(round(moments.ion.ppar[midpoint,1,1]*2/(
+                            moments.ion.dens[midpoint,1,1]), sigdigits = 4)), 7), "\n")
+                        end
```
johnomotani, line 2125: Maybe better to put this in the printouts that go with the 'moments' output, as sometimes that might be written more often?

johnomotani: > do you think the flag should exist in the timestepping section? Or better in the output section?

Good question, either seems sensible. The timestepping section is getting a bit long already, but we don't at the moment have a very convenient struct to put extra output options into to use for this feature, so keeping in timestepping seems the most sensible.

johnomotani (approved):

PR #345, opened 2025-04-15
Improve time-integration of PI controller integrals + external sources bugfixes

johnomotani:
When PI controllers were added, the original implementation of the error integral was a bit of a bodge, for simplicity. This PR updates the integral to be calculated correctly by the RK time-advance scheme. Should be more accurate (haven't tested this, but have checked that the updated one does push the temperature to the correct value in one case).

I think the original version effectively enhanced the 'I' factor of the PI scheme, by an amount related to the size of the coefficients in the RK scheme (which were ignored in the original implementation), so the new version might be slightly slower (I guess?) to converge to the target value unless you increase a bit the value for 'I' in the input file.

Bugfixes: a few factors of 1/density and 2 in branches that are either unused (so far...) or only affect very small terms.


johnomotani:
Updated to improve time-integration of PI controller integrals.

When PI controllers were added, the original implementation of the error integral was a bit of a bodge, for simplicity. This PR updates the integral to be calculated correctly by the RK time-advance scheme. Should be more accurate (haven't tested this, but have checked that the updated one does push the temperature to the correct value in one case).

I think the original version effectively enhanced the 'I' factor of the PI scheme, by an amount related to the size of the coefficients in the RK scheme (which were ignored in the original implementation), so the new version might be slightly slower (I guess?) to converge to the target value unless you increase a bit the value for 'I' in the input file.

Issue #351, opened 2025-04-23
Supplying inputs: complicated problems with sub-namelists


mrhardman:
I have a code-design issue in https://github.com/mabarnes/moment_kinetics/tree/implicit-fp-collisions-precon-debug which causes this test failure https://github.com/mabarnes/moment_kinetics/actions/runs/14595649141/job/40940781878.

I believe the issue is caused by restarting a simulation with the TOML from the output file, which is then inconsistent with the allowed inputs. The feature affected is the Fokker-Planck operator inputs, where the relevant code for understanding the error is here https://github.com/mabarnes/moment_kinetics/blob/0b0fa4b5d1c11ba6e43f253744f582d48dcecb09/moment_kinetics/src/fokker_planck.jl#L100-L179.

This function puts the input data into the struct `fkpl_collisions_input()`, but the data is read in two stages with two separate calls to `set_defaults_and_check_section!()`, because the current version of `set_defaults_and_check_section!()` cannot merge default `kwargs` recursively if one of the `kwargs` is an `OptionsDict()` type. This means that the default values for the nonlinear solver settings would not be picked up correctly if `nonlinear_solver_settings` was simply a default `kwarg` of the first call to `set_defaults_and_check_section!()`. Nonetheless, I want to pass this data around in `fkpl_collisions_input()` to avoid introducing extra data structures to pass around in the main code. The problem generated on restarting is that the restart TOML looks like it has picked up a namelist `[fokker_planck_collisions.nonlinear_solver]` that cannot be read by the `setup_fkpl_collisions_input()` function. Generated input below.

```
...
[fokker_planck_collisions]
use_fokker_planck = false
nuii = -1.0
frequency_option = "reference_parameters"
self_collisions = true
use_conserving_corrections = true
boundary_data_option = "direct_integration"
slowing_down_test = false
sd_density = 1.0
sd_temp = 0.01
sd_q = 1.0
sd_mi = 0.25
sd_me = 0.00013616557734204794
Zi = 1.0

    [fokker_planck_collisions.nonlinear_solver]
    rtol = 0.0
    atol = 1.0e-10
    linear_restart = 8
    linear_max_restarts = 1
    nonlinear_max_iterations = 20
    linear_rtol = 0.001
    linear_atol = 1.0
    preconditioner_update_interval = 300
    total_its_soft_limit = 50
    adi_precon_iterations = 1

[fokker_planck_collisions_nonlinear_solver]
rtol = 0.0
[...]

PR #352, opened 2025-04-23
Support subnamelists in one version of set_defaults_and_check_section!()

mrhardman:
This PR aims to update the version of `set_defaults_and_check_section!()` that returns a dictionary so that subnamelists can be supported in the TOML. Note that another version of  `set_defaults_and_check_section!()` at https://github.com/mabarnes/moment_kinetics/blob/322412d04710ab868aed0580dd8b22ac21b42bbc/moment_kinetics/src/input_structs.jl#L989-L1016 is not updated.

 - [ ] Consider if possible to use https://github.com/mabarnes/moment_kinetics/blob/322412d04710ab868aed0580dd8b22ac21b42bbc/moment_kinetics/src/utils.jl#L313-L329 or https://github.com/mabarnes/moment_kinetics/blob/322412d04710ab868aed0580dd8b22ac21b42bbc/moment_kinetics/src/utils.jl#L336-L345 to reduce code duplication.
 - [ ] Consider if possible or desirable to extend this support to https://github.com/mabarnes/moment_kinetics/blob/322412d04710ab868aed0580dd8b22ac21b42bbc/moment_kinetics/src/input_structs.jl#L989-L1016.

### DIFF
--- a/moment_kinetics/src/input_structs.jl
+++ b/moment_kinetics/src/input_structs.jl
@@ -927,8 +927,13 @@ function set_defaults_and_check_section!(options::OptionsDict, section_name::Str
         # `Base.get()` defined above.
         if isa(default_value, AbstractDict)
             # handle the case that the default_value is itself a dictionary from a subnamelist
+            if !(key in keys(section))
+                # create the subsection if it doesn't already exist
+                section[key] = OptionsDict()
+            end
+            subsection = section[key]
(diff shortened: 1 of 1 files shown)


mrhardman:
Closed as this method for supplying inputs seems inconsistent with the approach by @johnomotani for checking that all inputs are valid, as mentioned in #351. A new PR which should be consistent is #354.

PR #357, opened 2025-04-28
Implicit FP collisions sensible preconditioner defaults

mrhardman:
Set the defaults for functions involving implicit collisions so that the preconditioner is used by default. Note that this option is presently hard-coded for simulations using the main code rather than the test scripts. An option to switch this feature on or off could be included in this PR. But, it is not obvious where to put this input parameter for now.


johnomotani (approved):
Looks fine. I'd put the input setting in the `[fokker_planck_collisions]` section. Then either carry around in `collisions.fkpl` or use the `nl_solver_params.ion_fp_collisions.precon_type`, set it to something that can be compared at compile-time (i.e. has the information in the type) like `Val(:test_particle_precon)` similar to the kinetic electron options https://github.com/mabarnes/moment_kinetics/blob/0713657dea655ff0a76f1d2318efb25f9c204707/moment_kinetics/src/time_advance.jl#L459-L474

mrhardman:
> Looks fine. I'd put the input setting in the `[fokker_planck_collisions]` section. Then either carry around in `collisions.fkpl` or use the `nl_solver_params.ion_fp_collisions.precon_type`, set it to something that can be compared at compile-time (i.e. has the information in the type) like `Val(:test_particle_precon)` similar to the kinetic electron options
> 
> https://github.com/mabarnes/moment_kinetics/blob/0713657dea655ff0a76f1d2318efb25f9c204707/moment_kinetics/src/time_advance.jl#L459-L474

I opted for the simplest option for now, as the preconditioner is stored in the Fokker Planck arrays struct. We may move these options later if the preconditioner gets stored elsewhere, when or if we save it for all spatial points rather than recomputing it at each timestep. If the tests pass here, I will merge to #353.

PR #360, opened 2025-05-02
Optimize compilation of `get_variable()`; CI check for `makie_post_process()`

johnomotani:
Convert `String` variable names to `Symbol` in `get_variable()`. Speeds up compilation, I think by allowing the compiler to mostly optimise out the large if-elseif-... chain. On a test on my laptop, removed ~230s of compile time (a call to `makie_post_process()` in a fresh Julia session that makes just about every plot for a short simulation reduced in total run time from 681s (94% compile time) to 452s (91% compile time). Recompiling (using Revise.jl) when editing the `get_variable()` source code seemed a _lot_ quicker as well. Unfortunately this change makes `get_variable()` a little bit uglier, but I think it's worth it for the speed-up.

Also add a CI check for `makie_post_process()`. This generates a low-resolution output file with 2 output steps containing ions, electrons, and neutrals, then runs `makie_post_process()` on that output using a post-processing input file that switches on (just about) everything. This test only checks that there are no errors, not that the right output is produced.


johnomotani:
Merging this to master now. @LucasMontoya4 it might be worthwhile for you to update to the latest master to get this, it should speed up plotting a bit.

PR #361, opened 2025-05-02
Convert 'anyv' parallelisation to 'anysv'

### LINKED ISSUE #356: Change `anyv` region parallelisation to support implicit cross-species collisions in the future
The `anyv` region currently parallelises `z`, `r`, and `s` first, before allowing the user to change the local parallelisation of `vpa` and `vperp`. To make it possible to treat cross-species collisions, where and implicit solve must update an array of shape `pdf(vpa,vperp,s)` at each `z` and `r`, the species parallelisation needs to be treated on the same level as `vpa` and `vperp`.

To effect this change, (I think) we would need to update the list of parallelisation macros from `looping.jl` and the list of region macros from `communication.jl`. We would need to update the existing Fokker-Planck source to parallelise (or not) over species as appropriate to the operators in `fokker_planck.jl` and `fokker_planck_calculus.jl`. Since multispecies operation and cross-species collisions are not yet supported, the initial change could be the trivial one, perhaps changing code from the form

```
@begin_s_r_z_anyv_region()
    # parallelise over s, z, r, keep vpa, vperp flexible
    @loop_s_r_z is ir iz begin
         # call to single species self operator
         fokker_planck_self_collision_operator_weak_form!(pdf[:,:,is,ir,iz],pdf[:,:,is,ir,iz],...)
    end
end
``` 
to
```
@begin_r_z_anyv_region()
     # parallelise over z, r, keep s, vpa, vperp flexible
    @loop_r_z is ir iz begin
        #call to single species self operator
       is = 1
       fokker_planck_self_collision_operator_weak_form!(pdf[:,:,is,ir,iz],pdf[:,:,is,ir,iz],...)
    end
end
```
Presumably the available `anyv` region macros will go from
```
@begin_anyv_vperp_vpa_region()
@begin_anyv_vperp_region()
@begin_anyv_vpa_region()
@begin_anyv_region()
```
to 
```
@begin_anyv_s_vperp_vpa_region()
@begin_anyv_s_vperp_region()
@begin_anyv_s_vpa_region()
@begin_anyv_vperp_vpa_region()
@begin_anyv_vperp_region()
@begin_anyv_vpa_region()
@begin_anyv_s_region()
@begin_anyv_region()

```
and we will have to update `fokker_planck.jl` and `fokker_planck_calculus.jl`, and `nonlinear_solvers.jl` etc to use the appropriate region type for cases there there are multiple species and we want to parallelise over `s`. Otherwise, the same set of regions exist in the new as the old formulation, and the pseudocode change above should simply force `s` to be local in the Fokker Planck functions.

johnomotani:
Anticipate that when calculating multi-species collisions, the collision operators for every species should be calculated together, to be able to re-use parts of the computation. Therefore rather than 'anyv' regions inside an outer loop over species, r, and z, we want to have an outer loop over r and z, containing inner loops over both species and velocity space - call this an 'anysv' region (for any 's'pecies or 'v'elocity).

Resolves #356.

### DIFF
--- a/docs/src/developing.md
+++ b/docs/src/developing.md
@@ -231,8 +231,8 @@ that share an anysv subblock are all part of the `comm_anysv_subblock[]`
 communicator (which is a subset of the processes in the full block, whose
 communicator is `comm_block[]`).
 
-See also notes on debugging the 'anyv' parallelisation: [Collision operator and
-'anysv' region](@ref).
+See also notes on debugging the 'anysv' parallelisation: [Collision operator
+and 'anysv' region](@ref).
 
 ## Bounds checking
 
--- a/docs/src/developing.md
+++ b/docs/src/developing.md
@@ -195,43 +195,44 @@ Internally, when the `@begin_*_region()` macros need to change the region type (
 
 If for some reason it is necessary to synchronize explicitly, not by using an `@begin_*_region()` call, use the `@_block_synchronize()` macro. This calls `_block_synchronize()`, but when debugging can also pass in some information (a hash of the file an line number) about the calling site that is used for consistency checking.
 
-### Collision operator and `anyv` region
+### Collision operator and `anysv` region
 
 The Fokker-Planck collision operator requires a special approach to
-shared-memory parallelisation. There is an outer loop over spatial points (and
-potentially over species). Inside that outer loop there are operations that can
-benefit from parallelisation over $v_{\perp}$, or over $v_{\parallel}$, or over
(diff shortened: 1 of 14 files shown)


johnomotani:
@mrhardman for now this PR does not include any loops over species, and just hard-codes `is = 1` in the collision operator call https://github.com/mabarnes/moment_kinetics/blob/7cac2075cb8d2a2725281ef62f1bc36e507e2af2/moment_kinetics/src/fokker_planck.jl#L348
I'll leave it up to you to put the species-loops in the right places, and merge this with #353.

PR #363, opened 2025-05-08
Implicit fp collisions merge master

mrhardman:
An attempt to merge master with #322 into #353. Due to an update of the tests for Fokker Planck collisions to use a beam initial condition in #353, it was found that the `conserving_corrections!()` function required a modification to accommodate the change definition of the normalised heat flux `qpar`. With this change, appropriate normalisation changes to the expected results, and relaxing the tolerance from `atol=1.0e-14` to `2.0e-14`, all tests now pass apart from the explicit test of the value of `qpar` itself at the final time. A test of `conserving_corrections!()` is added to check the functionality of this function when the expected modifications are not small, and the moment `qpar` is nonzero.

In the  very low resolution Fokker Planck time revolution tests the heat flux `qpar` is nonzero, however if resolution is increased, this heat flux should in fact tend to 0 due to the symmetry of the initial beam distribution around `vpa0`. I have updated the test script `ImplicitCollisionsTest.jl` to permit a truly nonzero heat flux in the initial condition by summing two beam distributions and confirmed a higher resolution that this does indeed relax to 0, as expected in a Maxwellian, see below.
```
julia> include("test_scripts/ImplicitCollisionsTest.jl")
julia> test_implicit_collisions_wrapper(ntime=100, delta_t=1.0, nonlinear_max_iterations=20, atol=1.0e-14, test_numerical_conserving_terms=true, ngrid=5, nelement_vpa=32, nelement_vperp=16, Lvpa=8.0*sqrt(2), Lvperp=4.0*sqrt(2), vth0=0.5, vpa0 = 1.0*sqrt(2), vperp0=1.0, test_particle_preconditioner=true, zbeam=0.6)
begin elliptic operator assignment   11:36:47
finished elliptic operator constructor assignment   11:36:47
finished LU decomposition initialisation   11:36:48
finished YY array calculation   11:36:48
it = 0 time: 0.0
maximum(F_err): 0.11847061255465925 L2(F_err): 0.005158924379510866
dens: 0.999999999999998
upar: 0.8838834764831854
vth: 1.1455499890668743
ppar: 0.5937500000347852
qpar: -0.0828640759203
rmom: 4.792956617613492
...
it = 100 time: 100.0
maximum(F_err): 6.116942020961691e-8 L2(F_err): 1.9870045105547357e-9
dens: 1.0000000000000004
upar: 0.8838834764831875
vth: 1.1455499890668757
ppar: 0.6561423514585203
qpar: -1.2483559912423522e-7
[...]

PR #365, opened 2025-05-09
Fix of calculation of qpar for run time diagnostic

mrhardman:
This is a stub PR to demonstrate that fixing the calculation of `qpar` in run-time diagnostics for 2V requires us to update the tests in #363. We do not need to merge this PR if tests pass, instead, we should update #363. The calculation of `qpar` here is not optimal in 2V cases because a copy of `ff` is made before every call to `get_qpar()`.


johnomotani:
This looks like the right test. Hopefully this fix makes #363 consistent! :crossed_fingers:

mrhardman:
Closed as the tests here passed, permitting updated tests in #363 to be merged to #353.

PR #369, opened 2025-05-21
Allow launching multiple consecutive runs with `run_moment_kinetics.jl`

johnomotani:
Modifies the handling of command line arguments so that multiple input files can be passed to the `run_moment_kinetics.jl` script, to be run one after the other. The motivation is to avoid some setup/compile time when running MPI runs on a local machine, as it is inconvenient to run MPI runs interactively. I.e. you can now do something like
```bash
mpirun -np 32 bin/julia -O3 --project run_moment_kinetics examples/foo/input1.toml examples/foo/input2.toml examples/foo/input3.toml
```

Changes the syntax for restarting using `run_moment_kinetics.jl` - now the file to restart from has to be passed as the argument of the `--restartfile` flag, rather than as the second positional argument, because now the second positional argument might be another input file.

### DIFF
--- a/machines/archer/jobscript-restart.template
+++ b/machines/archer/jobscript-restart.template
@@ -22,6 +22,6 @@ export SRUN_CPUS_PER_TASK=$SLURM_CPUS_PER_TASK
 
 echo "running INPUTFILE $(date)"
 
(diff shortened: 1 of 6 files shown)


johnomotani:
Don't think this is super useful. We can reopen if there is a need for this feature.

PR #374, opened 2025-06-05
Add moment kinetic 2V formalism to code

LucasMontoya4:
Currently, one test (Harrison Thompson Chebyshev split 2) is failing - this may be because of changed integration routines for the moment constraints, or one of the interpolation changes - but I'm not sure yet.

Have written tests that run in 2V with no evolution and with evolution of each moment combination that's currently supported. Quantitatively when you look at the graphs of density, temperature etc. for the four combinations, just like in 1V, they are very close together, with the exception of (as is the case in 1V) the electric field at the boundaries.


johnomotani (changes requested):
Looks good. One small optimization I'd like to be added to this PR before merging.

johnomotani (approved):
:tada:

PR #380, opened 2025-07-08
Add ExB drift, support 2D geometry in moment-kinetic implementation

johnomotani:
Adds $E \times B$ drift and $B_z/B$ factors as in the TN-07 Excalibur report "2D drift kinetic model with wall boundary conditions" to the moment-kinetic implementation. Should support both 1V and 2V version, but not tested yet.

Adds option for spatially-varying temperature in source terms.

Adds check for CFL limit due to r-advection and vperp-advection in adaptive timestepping.

### DIFF
--- a/.github/workflows/test_scripts.yml
+++ b/.github/workflows/test_scripts.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macOS-latest]
       fail-fast: false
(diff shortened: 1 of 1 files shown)


johnomotani:
Merging now to make $B_z/B$ factors available in moment-kinetic code, and to minimise future merge conflicts. Inputs and tests to follow in a future PR.

PR #386, opened 2025-07-15
Moment kinetic Fokker Planck collisions

mrhardman:
This PR aims to bring support for Fokker-Planck collisions when moments are evolved separately from the pdf.

According to the documentation, the forms for the collision operator appearing in the kinetic equation when splitting the moments are as follows:
 - `evolve_density=true` https://mabarnes.github.io/moment_kinetics/dev/moment_kinetic_equations/#Separate-n_i 
```math
\frac{1}{n_i} C_ii[n_i F_i, n_i F_i] = n_i C_ii[ F_i,  F_i]
```
 - `evolve_upar=true` and  `evolve_density=true` https://mabarnes.github.io/moment_kinetics/dev/moment_kinetic_equations/#Separate-n_i-and-u_{i\\parallel} -- same as  `evolve_density=true` above.
 - `evolve_p = true` and `evolve_upar=true` and  `evolve_density=true` https://mabarnes.github.io/moment_kinetics/dev/moment_kinetic_equations/#ion_full_moment_kinetic_equation
 ```math
\frac{v_{Ti}^3}{n_i} C_ii[\frac{n_i}{v_{Ti}^3} F_i, \frac{n_i}{v_{Ti}^3} F_i] = \frac{n_i}{v_{Ti}^3} C_ii[ F_i,  F_i]
```
We include these factors for in the function https://github.com/mabarnes/moment_kinetics/blob/0e761f6260dfa4b5bf42bac3dc675be8cfa40e2b/moment_kinetics/src/fokker_planck.jl#L818-L836.

We do not support moment kinetic features in cross collisions or in https://github.com/mabarnes/moment_kinetics/blob/0e761f6260dfa4b5bf42bac3dc675be8cfa40e2b/moment_kinetics/src/fokker_planck.jl#L280-L290.

To do:
 - [] Add tests of moment-kinetic normalisation factors.

### DIFF
--- a/moment_kinetics/src/fokker_planck.jl
+++ b/moment_kinetics/src/fokker_planck.jl
@@ -352,7 +352,9 @@ end
 Function for advancing with the explicit, weak-form, self-collision operator.
 """
 @timeit global_timer explicit_fokker_planck_collisions_weak_form!(
-                         pdf_out, pdf_in, dSdt, composition, collisions, dt,
+                         pdf_out, pdf_in,
+                         dSdt, density, vth, evolve_p, evolve_density,
+                         composition, collisions, dt,
                          fkpl_arrays::fokkerplanck_weakform_arrays_struct, r, z, vperp,
                          vpa, vperp_spectral, vpa_spectral, scratch_dummy;
                          diagnose_entropy_production=false) = begin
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by johnomotani** [comment]
I'm working on some tests so we can get this merged.

**Review by johnomotani — APPROVED** [review]
Looks good to me.

When I was adding tests, the default resolution in fokker_planck_time_evolution_tests.jl was not high enough to make moment-kinetic and full-f versions agree closely enough to be a nice test (I think the differences were >1%). With higher velocity-space resolution, the errors drop down to small values (so implementation seems correct), but the test is too slow for CI. To get around this, I've saved expected results separately for each supported moment-kinetic option for the CI test. I also saved high-resolution expected output (from the full-f version), with appropriate tolerances that the moment-kinetic versions should agree within - the high-resolution test can be run manually by doing something like
```julia
julia> include("moment_kinetics/test/fokker_planck_time_evolution_tests.jl")
julia> FokkerPlanckTimeEvolutionTests.runtests(; highres=true)
```
```

PR #387, opened 2025-07-17
Improve handling of cut-off points in moment-kinetic ion wall bc

johnomotani:
In the moment-kinetic approach, the point where v_parallel=0 moves on the w_parallel grid as upar and vth change. Previously, when the v_parallel=0 point passed a grid point, the shape function on that grid point would jump from non-zero to zero. This jump caused a jump in the (numerical) moments of the shape function, a jump in the moment corrections, and a jump in the parallel heat flux. The jump in the heat flux then feeds back into the kinetic equation and moment equations, causing numerical instability that could severely restrict the timestep when the velocity space resolution is not high enough. In particular if the v_parallel=0 point happened to land very close to a grid point in the steady state of a simulation, it could happen that this jumping would make the v_parallel=0 point flip back and forth from one side of the grid point to the other, preventing the simulation from reaching a completely steady state.

The issue can be fixed by ensuring that the value of the shape function on the final non-zero grid point goes to zero in some continuous way as the v_parallel=0 point approaches the grid point. This PR ensures this by limiting the value of the shape function on that grid point to be at maximum the linear fit between the second-last non-zero grid point and zero at the v_parallel=0 point. Using a maximum cap like this, rather than for example decreasing by some scale factor, ensures that repeated application of the boundary condition with the same inputs produces exactly the same output - this property has been found previously to be important for numerical stability.


johnomotani:
Note the important part of this PR is https://github.com/mabarnes/moment_kinetics/pull/387/files#diff-e55db06c75c6cd18763d8c571f4f2fc02f3c31b0ee1d3afc24cf497e850f99a3R1613-R1673. The remaining changes are just reverting a modification that #380 made to `vpagrid_to_dzdt()`, and renaming the function to `vpagrid_to_vpa()` as it actually converts $w_\parallel$ (`vpa.grid`) to $v_\parallel$, not to $\dot z$.

Issue #220, opened 2024-05-29
Support Chebyshev Grids for collision operator and other weak form operations


mrhardman:
Currently the collision operator is only implemented for Gauss Legendre grids. This could be overcome if the setup for the various underlying matrices involving integrals of produces of lagrange polynomials was added for the Chebyshev grid. This take could be completed relatively straightforwardly if we can provide and option to compute these matrices using the explicit interpolation formula for the Lagrange polynomials and their derivatives, in terms of arbitrary grid locations. (I believe @johnomotani already started work on this some time ago).

Currently the reason for avoiding this approach in the Gauss Legendre grids and instead using a formulation in terms of Legendre polynomials was to gain understanding of the weak-form method with definite testable results provided by http://dx.doi.org/10.1016/j.jcp.2014.12.012.

PR #388, opened 2025-07-22
Use split packages for Fokker Planck collision operator

mrhardman:
Attempt to use [FokkerPlanck](https://github.com/moment-kinetics/FokkerPlanck) to provide the collision operator. This PR will hopefully reduce code volume in various modules without losing functionality.

The main features of this PR are to:
 - Replace functions related to Fokker Planck collisions with those from the [external package](https://github.com/moment-kinetics/FokkerPlanck)
 - Update tests where changes were required due to minor differences in numerical implementation in the external package.
 - Remove test scripts and other code that are now featured in the external package.
 - Provide an interface to Fokker Planck collisions that will function for any element-based tensor product grid in moment kinetics (Chebyshev and Gauss-Legendre grids should both function).
 - Use [LagrangePolynomials](https://github.com/moment-kinetics/LagrangePolynomials) for methods requiring Lagrange polynomials.
 - Use [FiniteElementMatrices](https://github.com/moment-kinetics/FiniteElementMatrices) where appropriate to provide elemental matrices for weak methods (localised only to the `gauss_legendre.jl` module for now, but other developers could now merge `gauss_legendre.jl` and `chebyshev.jl`, with some careful thought).

N.B.
 - This package removes support for velocity-space shared-memory MPI for collisions. This is argued to not be critical as the spatial parallelisation is 1) retained and 2) embarrassingly parallel and 3) likely to require at least `ngrid*ngrid` cores on the local node to be fully parallelised. For `ngrid=5`, it is easy to imagine that we could break the domain into blocks of sized ~20 core shared-memory regions and only parallelise over space.

### DIFF
--- a/moment_kinetics/Project.toml
+++ b/moment_kinetics/Project.toml
@@ -17,7 +17,6 @@ HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
 Interpolations = "a98d9a8b-a2ab-59e6-89dd-64a1c18fca59"
 JacobianFreeNewtonKrylov = "ea74541f-29a8-443c-83cd-bcc24bd49708"
 LagrangePolynomials = "fb83b288-ce1a-45f4-939e-c34c34051e49"
-LegendrePolynomials = "3db4a2ba-fc88-11e8-3e01-49c72059a882"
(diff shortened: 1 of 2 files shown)


mrhardman:
Closes #220.

PR #412, opened 2025-10-29
Better chunk sizes when writing to HDF5

johnomotani:
Previously, the 'chunk size' for HDF5 writing was always set to the total size of the array to be written from a single distributed-MPI subdomain. However, for large arrays this can cause problems (very slow writing or errors depending on the MPI library used - thanks @LucasMontoya4 for digging into this!). Also, when reading slices of large arrays back in, this choice would be non-optimal as to load any piece of data, HDF5 has to load the entire chunk containing it - therefore when only a slice needs loading, it is best if the dimension being sliced is chunked into small parts. According to ARCHER2 documentation, 1MiB (2^20 bytes) is a good block size to write on that machine, so that seems like a sensible default. With this PR, the chunk size is the array size if the array is smaller than 8MiB (bigger than 1MiB to ensure there are not almost-empty chunks being written, e.g. if the array size was 1.0001MiB), or otherwise the chunk size is 1MiB exactly, with the dimensions divided up as evenly as possible so that there is no preferred or non-preferred (in terms of read performance) dimension to slice.

Also update HDF5 version in machine_setup to version 1.14.6 and save configure and make logs from HDF5 build. Note that after updating the HDF5 library install, it is necessary to ensure that the HDF5.jl package precompiles again, otherwise there may be errors when writing String attributes (https://github.com/JuliaIO/HDF5.jl/issues/1210).

### DIFF
--- a/moment_kinetics/src/file_io.jl
+++ b/moment_kinetics/src/file_io.jl
@@ -1344,7 +1344,7 @@ function define_dynamic_ion_moment_variables!(fid, n_ion_species, r::coordinate,
         evolve_p, write_error_diagnostics, write_steady_state_diagnostics)
 
     dynamic = get_group(fid, "dynamic_data")
-    ion_species_coord = (name="ion_species", n=n_ion_species)
+    ion_species_coord = (name="ion_species", n=n_ion_species, nrank=1)
 
     # io_density is the handle for the ion particle density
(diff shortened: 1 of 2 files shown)


johnomotani:
We could allow the 'target' chunk size to be a user input (at least for powers of 2 other than 20) in case a value other than 1MiB is optimal), but I/O performance is not usually important for our simulations, so it doesn't seem necessary to implement this for now.

PR #413, opened 2025-10-30
1D ITG tests and 1D arc length field line trapping geometries

LucasMontoya4:


moment_kinetics/src/load_data.jl
```diff
@@ -3712,10 +3712,6 @@ function get_run_info_no_setup(run_dir::Union{AbstractString,Tuple{AbstractStrin
         vz_chunk_size = 1
     end
 
-    looping.setup_loop_ranges!(block_rank[], block_size[]; s=composition.n_ion_species,
-                               sn=composition.n_neutral_species, r=r.n, z=z.n,
-                               vperp=vperp.n, vpa=vpa.n, vzeta=vzeta.n, vr=vr.n, vz=vz.n)
-
```
johnomotani: I've forgotten now exactly why we did this - I feel like there was a good reason. Just adding a comment here so it's easy to find the change if it turns out to cause some problem in the future.

johnomotani: Ah, it's because we moved the `setup_loop_ranges!()` call into `mk_input()`, so it's already been done!

moment_kinetics/src/r_advection.jl
```diff
@@ -47,6 +48,43 @@ do a single stage time advance (potentially as part of a multi-stage RK scheme)
...
+    for is in 1:1
+        for iz in 1:z.n
+            for ivperp in 1:vperp.n
+                for ivpa in 1:vpa.n
+                    f_out[ivpa,ivperp,iz,1,is] += B_factor * Ez[ivperp,iz,1,is] * f[ivpa,ivperp,iz,1,is]/L_T *
+                                                    (1/2 * (vperp.grid[ivperp]^2 + vpa.grid[ivpa]^2)/T[iz,1,is] - 3/2) * dt
+                end
+            end
+        end
+    end
```
johnomotani: This will only work in serial. Could wrap in a `@begin_serial_region(); @serial_region begin...end`, but probably better/easier to just use the `@loop_*` macro for the loop. If you move the calculation of `T` inside the species and z loops, and just calculate it as a scalar variable for that `is` and `iz`, then there would be no need to allocate/fill an array for `T`.

johnomotani (approved):

johnomotani:
> The ubuntu tests seem to be failing the `ITG_1D` tests. They pass on my Mac, but the evaluation fails on the ubuntu machines, with seemingly around `1.0e-7` being the point at which phi is off for ubuntu vs macOS. Not sure what the cause is here, hopefully I can reproduce this error on my own Linux machine

Ah, the problem is probably just that the tolerances are a bit too tight then. The rounding-error differences are usually bigger when changing operating system (i.e. Linux<->macOS) than when changing machine on the same system/hardware. There's a different type of CPU, different compilers, etc. So as long as the error still seems 'small' it should be fine to just make sure the tolerance is large enough that both CI servers pass with a little bit of a margin (I usually go for setting the tolerance so that very roughly the actual error at the moment is half (more like 50% than 90% or 10%...) what the tolerance would allow to pass, to give a bit of margin for error if libraries/compilers/etc. change in future). I often have to do this when I generate expected results on Linux and then the CI job fails on macOS.

PR #418, opened 2025-11-05
Move preconditioners setup out of `setup_nonlinear_solve()`

johnomotani:
The exact set of things that need to go into the `preconditioners` Tuple is problem and solver specific, so makes more sense to define this in the same place as the preconditioners themselves. Allows the nonlinear_solvers module to be more generic, which will make it easier to split into a separate package in future.

Bugfix for setup when there are multiple copies of the preconditioner (e.g. for different points in r).


mrhardman:
In case it is helpful, I note that I came to a similar conclusion when splitting the JFNK solver for FokkerPlanck.jl https://github.com/moment-kinetics/JacobianFreeNewtonKrylov. There I assume that the preconditioner function is defined where the `newton_solve!()` function is called, and that the user can extend the definition of the various internal functions as appropriate (see here https://github.com/moment-kinetics/FokkerPlanck/blob/main/src/fokker_planck_nonlinear_solvers.jl).

PR #455, opened 2026-06-04
Improve `coll_krook` second derivative

johnomotani:
Slightly hacky, but only expect to use this for the Krook-collisional fluid closure. Should help to prevent spikes at z-direction element boundaries in the ion temperature (in coll_krook simulations).

Also puts the update of `moments.ion.temp` in a more standard place, so that the ion temperature is always available (this could be used to simplify code in several places, but this has not been done yet).


johnomotani:
@LucasMontoya4 - tests passed, should we merge this now?

LucasMontoya4:
@johnomotani looks fine to me. Do you think it's alright to have my Zenodo repo for the first paper still point to the previous version of master, so that the simulations can stay as the old ones?

johnomotani:
> Do you think it's alright to have my Zenodo repo for the first paper still point to the previous version of master, so that the simulations can stay as the old ones?

Yes that's fine. The data provenance should point to whatever version (or versions, if necessary!) you used.

